PR #43, opened 2023-08-31
Fix guest example target path

skunert:
The target path in the build script was pointing to the top level target dir.


koute:
Thanks!

Since you're running this script, did you get an rv32e toolchain built?

skunert:
Yes, I was using your instructions here: https://github.com/koute/rustc-rv32e

PR #44, opened 2023-09-01
Disassembler cli

f-gate:
Hi there, cool project. 
Thought id give this a go as its easy enough for my first open source contribution.

Solves: https://github.com/koute/polkavm/issues/34


koute (changes requested):
Thanks for the PR!

Also, a question: did you run `rustfmt`? If not please run `cargo fmt --all`. (Sorry, I haven't yet set up the CI so this is not yet automatically verified.)

crates/polkavm/src/disassembler.rs
```diff
@@ -0,0 +1,9 @@
+use crate::Module;
+use polkavm_common::program::RawInstruction;
+
+pub struct Disassembler;
+impl<'a> Disassembler {
+    pub fn from_module(module: &'a Module) -> impl Iterator<Item = &RawInstruction> {
+        module.instructions().iter()
+    }
+}
```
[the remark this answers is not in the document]
f-gate: I attempted to use `module.instructions()` however, the visibility is only pub(crate). I thought this would be less invasive than  changing the visibility of the `module.instructions()` :).
```
 pub(crate) fn instructions(&self) -> &[RawInstruction] {
        &self.0.instructions
    }
```
Unless there is something im missing here.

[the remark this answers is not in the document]
koute: You don't need to go through a `Module`. You can just parse the blob into a `ProgramBlob` and then call [`.instructions`](https://docs.rs/polkavm-common/0.1.1/polkavm_common/program/struct.ProgramBlob.html#method.instructions) on it, which is public.

tools/polkatool/src/main.rs
```diff
@@ -13,6 +13,16 @@ enum Args {
         /// The input file.
         input: PathBuf,
     },
+
+    /// Disassembles a .polkavmm blob into its human-readable assembly.
```
koute, line 17:
```suggestion
    /// Disassembles a .polkavm blob into its human-readable assembly.
```

Typo.

tools/polkatool/src/main.rs
```diff
@@ -13,6 +13,16 @@ enum Args {
...
+    Disassemble {
+        /// The output file.
+        #[clap(short = 'o', long)]
+        output: PathBuf,
```
koute, line 21:
```suggestion
        output: Option<PathBuf>,
```

Wasn't made optional. (:

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,42 @@ fn main() {
...
+                    Err(error) => {
+                        eprintln!(
+                            "ERROR: failed to parse raw instruction from blob. {:?}: {}. nth:{} ",
+                            input, error, nth
+                        );
+                        std::process::exit(1);
+                    }
+                })
+                .collect::<Vec<String>>()
+                .join("\n");
```
koute, line 84: Open the file normally and turn this into a loop, instead of turning everyone into strings first.

For example, something like this:

```
let fp = match File::open(path) { .. };
let mut fp = BufWriter::new(fp);
for (nth_instruction, instruction) in blob.instructions.enumerate() {
    let instruction = match instruction { .. };
    writeln!(&mut fp, ..).unwrap();
}

match fp.flush() { .. }
```

koute (changes requested):

tools/polkatool/src/main.rs
f-gate: oops haha

tools/polkatool/src/main.rs
f-gate: Awesome! thanks for this, much cleaner and simpler.

tools/polkatool/Cargo.toml
```diff
@@ -10,3 +10,4 @@ description = "A swiss knife toolkit for PolkaVM"
 [dependencies]
 clap = { workspace = true, features = ["derive"] }
 polkavm-linker = { workspace = true }
+polkavm = { workspace = true }
```
koute, line 13:
```suggestion
```

You forgot to remove this. (:

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,55 @@ fn main() {
...
+            let blob = match polkavm::ProgramBlob::parse(&data[..]) {
+                Ok(b) => b,
+                Err(error) => {
+                    eprintln!("ERROR: failed to parse the raw data into a blob {:?}: {}", input, error);
+                    std::process::exit(1);
+                }
+            };
+
+            let std_out = std::io::stdout();
+            let mut bw: Box<dyn Write> = match output {
```
koute, line 72: This doesn't need to be boxed.

The easiest way to deal with it is to just move the body which does the dissasembly into a separate function, e.g. something like this:

```rust
fn disassemble_into(blob: &ProgramBlob, out: &mut dyn Write) {
    // ...
}

// ...
if let Some(path) = output {
    let mut fp = ...;
    disassemble_into(&blob, &mut fp);
} else {
    let mut fp = BufWriter::new(std::io::stdout());
    disassemble_into(&blob, &mut fp);
}
```

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,55 @@ fn main() {
...
+                            "ERROR: failed to parse raw instruction from blob. {:?}: {}. nth:{} ",
+                            input, error, nth_instruction
+                        );
+                        std::process::exit(1);
+                    }
+                };
+                if let Err(error) = writeln!(&mut bw, "{}", instruction) {
+                    eprintln!("ERROR: failed to write instruction from buffer. {:?} {}.", error, instruction);
+                    std::process::exit(1);
+                }
```
koute, line 100: Just use `writeln!` directly instead of doing a `format!` first. (:

Also you can put the variables inline into the string itself.

This works:

```rust
writeln!(&mut fp, "{nth_instruction}: {instruction}")
```

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,55 @@ fn main() {
...
+            }
+            bw.flush().unwrap();
```
koute, line 102: This can fail with a write error, so don't `unwrap` it.

koute (changes requested):

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,63 @@ fn main() {
...
+                eprintln!(
+                    "ERROR: failed to parse raw instruction from blob. {}. nth:{} ",
+                    error, nth_instruction
+                );
+                std::process::exit(1);
+            }
+        };
+    }
+    if let Err(error) = writer.flush() {
+        eprintln!("ERROR: failed to flush buffer writer. {}.", error);
```
koute, line 110:
```suggestion
        eprintln!("ERROR: failed to write to output: {}", error);
```

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,63 @@ fn main() {
...
+            Ok(instruction) => {
+                if let Err(error) = writeln!(&mut writer, "{nth_instruction}: {instruction}") {
+                    eprintln!("ERROR: failed to write instruction from buffer. {:?} {}.", error, instruction);
+                    std::process::exit(1);
+                }
+            }
+            Err(error) => {
+                eprintln!(
+                    "ERROR: failed to parse raw instruction from blob. {}. nth:{} ",
+                    error, nth_instruction
```
koute, line 103:
```suggestion
                    "ERROR: failed to parse instruction #{}: {}", nth_instruction, error
```

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,63 @@ fn main() {
...
+        }
+    }
+}
+
+fn disassemble_into(blob: &polkavm_linker::ProgramBlob, mut writer: impl Write) {
+    for (nth_instruction, maybe_instruction) in blob.instructions().enumerate() {
+        let instruction = match maybe_instruction {
+            Ok(instruction) => {
+                if let Err(error) = writeln!(&mut writer, "{nth_instruction}: {instruction}") {
+                    eprintln!("ERROR: failed to write instruction from buffer. {:?} {}.", error, instruction);
```
koute, line 96:
```suggestion
                    eprintln!("ERROR: failed to write to output: {}", error);
```

tools/polkatool/src/main.rs
```diff
@@ -41,5 +51,63 @@ fn main() {
...
+                Ok(data) => data,
+                Err(error) => {
+                    eprintln!("ERROR: failed to read {:?}: {}", input, error);
+                    std::process::exit(1);
+                }
+            };
+            let blob = match polkavm_linker::ProgramBlob::parse(&data[..]) {
+                Ok(b) => b,
+                Err(error) => {
+                    eprintln!("ERROR: failed to parse the raw data into a blob {:?}: {}", input, error);
```
koute, line 66:
```suggestion
                    eprintln!("ERROR: failed to parse {:?}: {}", input, error);
```

tools/polkatool/Cargo.toml
```diff
@@ -9,4 +9,4 @@ description = "A swiss knife toolkit for PolkaVM"
 
 [dependencies]
 clap = { workspace = true, features = ["derive"] }
-polkavm-linker = { workspace = true }
+polkavm-linker = { workspace = true }
```
koute, line 12: Could you please restore this newline? (: (Sorry!)

koute (changes requested):

koute:
Almost good to go!

f-gate:
thanks for the awesome review, if it wasnt too much hassle for you ill try picking up some other pieces.

tools/polkatool/Cargo.toml
koute: I meant the newline at the very end of the file, not *before* this line. (:

But it's fine; looks like I was able to fix it myself.

koute:
Okay, can you fix clippy and rustfmt?

koute:
Thanks!

> if it wasnt too much hassle for you ill try picking up some other pieces

That's fine, but please do consult with me before you pick anything substantial. Things are still very much in flux so not everything is appropriate for people to pick. (:

PR #50, opened 2023-09-22
Quick typos fixes

pgherveou:


koute:
Thanks!

Can you remove the changes from `program_from_elf.rs`? (I'm currently significantly refactoring/rewriting that part of the code, I just haven't pushed this yet as it's not yet finished. So it's not necessary to fix those as it'll just create patch conflicts.)

pgherveou:
done

PR #55, opened 2023-10-19
Update README.md to fix incorrect description

starit:
"Other than" means not included. So it means it does not support amd64 and arm64(also does not support Windows, Linux and macOS)

I guess the initial purpose is to express it supports those architectures and OSs. So it should use "in addition to"  or "no other than".

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ PolkaVM is a general purpose user-level RISC-V based virtual machine.
 ## Non-goals
 
 - System level emulation. This VM will never be able to run a normal operating system.
-- Full support for architectures other than amd64 (also known as x86_64) and aarch64 (also known as arm64). Anything else will run in an interpreted mode.
-- Full support for operating systems other than Linux, macOS and Windows. On any other OS the VM will run in an interpreted mode.
+- Full support for architectures in addition to amd64 (also known as x86_64) and aarch64 (also known as arm64). Anything else will run in an interpreted mode.
+- Full support for operating systems in addition to Linux, macOS and Windows. On any other OS the VM will run in an interpreted mode.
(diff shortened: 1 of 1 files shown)


koute:
I'm confused; to me it seems that the current usage is correct? "Full support for architectures other than amd64 and aarch64 [is a non-goal]" is exactly what I meant there.

starit:
> I'm confused; to me it seems that the current usage is correct? "Full support for architectures other than amd64 and aarch64 [is a non-goal]" is exactly what I meant there.

Do you mean that arm64 and amd64 are not supported by PolkaVM? Just to confirm...

koute:
> > I'm confused; to me it seems that the current usage is correct? "Full support for architectures other than amd64 and aarch64 [is a non-goal]" is exactly what I meant there.
> 
> Do you mean that arm64 and amd64 are not supported by PolkaVM? Just to confirm...

No; those are the only architectures which will be supported. The section is called "**non**-goals", which means that every bullet point is supposed to be taken as a negative, as in "these things are **not** the goals". In this case, "**no** full support for architectures other than amd64 and aarch64".

starit:
> > > I'm confused; to me it seems that the current usage is correct? "Full support for architectures other than amd64 and aarch64 [is a non-goal]" is exactly what I meant there.
> > 
> > 
> > Do you mean that arm64 and amd64 are not supported by PolkaVM? Just to confirm...
> 
> No; those are the only architectures which will be supported. The section is called "**non**-goals", which means that every bullet point is supposed to be taken as a negative, as in "these things are **not** the goals". In this case, "**no** full support for architectures other than amd64 and aarch64".

Understood. I thought we usually put something nice to have on the non-goals. Misunderstood. Thanks for the explanation.

PR #69, opened 2023-11-23
Use AsmMachine's Trace execution

mohanson:
Although both two writings can execute the virtual machine normally, the old writing is usually used for single-step debugging.


koute:
Ah, I see! Thanks!

Perhaps that function should be renamed to make this more obvious? (:

PR #100, opened 2024-02-21
Dynamic memory allocation support; refactoring

koute:
Fixes https://github.com/koute/polkavm/issues/52
Fixes https://github.com/koute/polkavm/issues/36
Reverts https://github.com/koute/polkavm/issues/80
Partially implements https://github.com/koute/polkavm/issues/72 (the `xtheadcondmov` extension)


crates/polkavm-linker/src/riscv.rs
```diff
@@ -400,6 +400,43 @@ fn test_bits() {
     assert_eq!(unbits(5, 10, 2048, 25), 0);
 }
 
+#[derive(Copy, Clone)]
+pub struct R(pub u32);
+
+// See chapter 19 of the RISC-V spec.
+pub const OPCODE_CUSTOM_0: u32 = 0b0001011;
```
nazar-pc, line 407: Inline assembly that is generated for imports uses this opcode as `0xb`.

I was initially searching for `0xb` in the code and wasn't able to find it. It'd be helpful to have a link between inline assembly and this constant.

Either by using `0xb` here (though I understand how bit representation might be valuable here) or maybe referencing constant directly (if possible) or switching from hex to binary number in inline assembly (I do not know if it is supported).

PR #109, opened 2024-04-30
Refactor of the disassembler out of `polkatool` to make it re-usable

xermicus:
Quick refactor of the disassembler out of `polkatool` and into a dedicated crate so I can re-use it in the YUL compiler.

Mostly a linear refactor as I tried to not change any logic and keep syntactic/style changes to a minimum.


koute:
Looks mostly good to me!

Hmm... disassembling pinky/doom might be a little bit of an overkill though, and those tests don't actually test the disassembly itself, just that it doesn't error out. Maybe instead you could dynamically assemble a simple program [like this](https://github.com/koute/polkavm/blob/master/crates/polkavm/src/tests.rs#L779), disassemble that, and also `assert_eq` the exact output of the disassembly?

xermicus:
Yeah I agree, I just wasn't aware of the super convenient program builder! Added a simple program with an explicit check against the disassembly.

I thought that pinky and doom probably contain a lot (most) instructions and if they don't panic and disassemble into an non empty string, the disassembler _likely_ works fine.. If you'd like me to remove them I can do that, however I still think it wouldn't harm to exercise those anyways :) WDYT?

koute:
> I thought that pinky and doom probably contain a lot (most) instructions and if they don't panic and disassemble into an non empty string, the disassembler _likely_ works fine.. If you'd like me to remove them I can do that, however I still think it wouldn't harm to exercise those anyways :) WDYT?

Well, they're probably unnecessary, but we can always easily remove them later so it's fine either way.

xermicus:
@koute fair enough, I've removed them.

koute:
Thanks!

PR #123, opened 2024-05-10
Prevent sections with no data from bloating up program blobs

koute:
Fixes https://github.com/koute/polkavm/issues/122


xermicus:
@koute thanks!

PR #127, opened 2024-05-15
update Wasmi in benchtool

Robbepop:
- Adds `lazy.unchecked` Wasmi configuration since that is used in pallet-contracts for calls
- Enables `no-hash-maps` feature since that is used in pallet-contracts


koute:
Thanks!

PR #134, opened 2024-06-24
Support for compressed instructions

### LINKED ISSUE #59: Support the C extension
- This has to only be handled in the linker.
- Functionally this shouldn't make a difference, but it will allow us to assume that code addresses are only 16-bit aligned, which will allow us to cut the size of the jump tables by half.

xermicus:
Closes #59


xermicus:
Added a relinked doom - runs like a charm on my M2 (compiler backend via rosetta)

PR #140, opened 2024-06-29
AppArmor issue on ubuntu 24

rosarp:
Added check for is_sandboxing_supported

### DIFF
--- a/crates/polkavm/src/sandbox/linux.rs
+++ b/crates/polkavm/src/sandbox/linux.rs
@@ -34,16 +34,55 @@ use crate::config::GasMeteringKind;
 use crate::shm_allocator::{ShmAllocator, ShmAllocation};
 
 pub struct GlobalState {
-    shared_memory: ShmAllocator
+    shared_memory: ShmAllocator,
+    is_sandboxing_enabled: bool,
 }
 
 impl GlobalState {
-    pub fn new(_config: &Config) -> Result<Self, Error> {
+    pub fn new(config: &Config) -> Result<Self, Error> {
+        let is_sandboxing_enabled = GlobalState::is_sandboxing_supported();
+        if !config.allow_insecure && !is_sandboxing_enabled  {
+            return Err(Error::from_str("failed to spawn new process"));
+        }
+
         Ok(GlobalState {
-            shared_memory: ShmAllocator::new()?
+            shared_memory: ShmAllocator::new()?,
+            is_sandboxing_enabled,
         })
     }
+
+    fn is_sandboxing_supported() -> bool {
+        let child_pid = unsafe { linux_raw::syscall!(linux_raw::SYS_clone, SANDBOX_FLAGS, 0, 0, 0, 0) };
+        if child_pid < 0 {
+            return false;
+        }
+
+        if !cfg!(polkavm_dev_debug_zygote) {
+            // Overwrite the hostname and domainname.
+            match linux_raw::sys_sethostname("localhost") {
+                Ok(()) => {},
+                Err(error) => {
+                    log::trace!("Failed to call sys_sethostname: {}", error);
+                    return false
+                },
+            }
+        }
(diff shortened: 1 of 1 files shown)


crates/polkavm/src/sandbox/linux.rs
```diff
@@ -34,17 +34,56 @@ use crate::config::GasMeteringKind;
...
+            match linux_raw::sys_sethostname("localhost") {
+                Ok(()) => {},
+                Err(error) => {
+                    log::trace!("Failed to call sys_sethostname: {}", error);
+                    return false
+                },
+            }
+        }
+        true
+    }
```
koute, line 71: This is incorrect as it will leave two processes running. (:

What the `SYS_clone` call does is that it duplicates the current process, and returns `0` in the copied process, and a non-zero value in the original process (basically the `syscall!(SYS_clone)` will return **twice**; this is similar to the `fork` function call, and if you've never used it I suggest you play with it first).

The way this should be implemented is roughly like this (pseudocode):

```
let sigset = Sigmask::block_all_signals()?;
let child_pid = sys_clone();
if child_pid < 0 {
    return false;
}
if child_pid == 0 {
    core::mem::forget(sigset);
    // We're in the child.
    let status = if linux_raw::sys_sethostname(...).is_ok() { 0 } else { 1 };
    linux_raw::sys_exit(status);
    linux_raw::abort();
} else {
    // We're in the host.
    sigset.unblock()?;
    let mut child = ChildProcess { pid: child_pid, pidfd: None };
    let status = child.wait_for_exit();
    return status == 0;
}
```

The idea here is to use the child process' exit status to communicate to the host process whether sandboxing can work. Getting an error message from the child to the host (as to why the sethostname failed) would also be nice, but we can leave that for another PR.

crates/polkavm/src/sandbox/linux.rs
```diff
@@ -34,17 +34,56 @@ use crate::config::GasMeteringKind;
...
+    shared_memory: ShmAllocator,
+    is_sandboxing_enabled: bool,
 }
 
 impl GlobalState {
-    pub fn new(_config: &Config) -> Result<Self, Error> {
+    pub fn new(config: &Config) -> Result<Self, Error> {
+        let is_sandboxing_enabled = GlobalState::is_sandboxing_supported();
+        if !config.allow_insecure && !is_sandboxing_enabled  {
+            return Err(Error::from_str("failed to spawn new process"));
```
koute, line 45: The error message could be better; maybe something like "sandboxing is not supported on your system; you can enable `set_allow_insecure`/`POLKAVM_ALLOW_INSECURE` to run with reduced sandboxing if you know what you're doing".

koute (changes requested):

crates/polkavm/src/sandbox/linux.rs
rosarp: Updated the error message.

crates/polkavm/src/sandbox/linux.rs
```diff
@@ -34,17 +34,56 @@ use crate::config::GasMeteringKind;
...
+                    return false
+                },
+            }
+        }
+        true
+    }
 }
 
+const SANDBOX_FLAGS: u32 =
+    if !cfg!(polkavm_dev_debug_zygote) {
```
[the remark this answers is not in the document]
rosarp: removed cfg here

crates/polkavm/src/sandbox/linux.rs
rosarp: allow_insecure is removed from config. updated this fn to handle child process.

crates/polkavm/src/sandbox/linux.rs
rosarp: since allow_insecure is removed, this message needs to be updated. Any suggestions?

PR #142, opened 2024-07-01
bump `object` crate dependency

xermicus:
Their latest version includes more RISC-V reloc consants which we'll need anyways.

### DIFF
--- a/crates/polkavm-linker/src/elf.rs
+++ b/crates/polkavm-linker/src/elf.rs
@@ -118,7 +118,7 @@ impl<'data, 'file> Symbol<'data, 'file> {
     }
 
     pub fn kind(&self) -> u8 {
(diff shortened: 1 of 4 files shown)


crates/polkavm-linker/src/elf.rs
```diff
@@ -132,23 +132,23 @@ pub struct Elf<'data> {
 impl<'data> Elf<'data> {
     pub fn parse(data: &'data [u8]) -> Result<Self, ProgramFromElfError> {
         let elf = ElfFile::parse(data)?;
-        if elf.raw_header().e_ident.data != object::elf::ELFDATA2LSB {
+        if elf.elf_header().e_ident.data != object::elf::ELFDATA2LSB {
```
xermicus, line 135: This is just to fix some deprecation warnings

PR #144, opened 2024-07-17
add handling of load_imm_and_jump_indirect

subotic:
Adds support for `load_imm_and_jump_indirect`:
  - short form: `a1 = 1234, jump [a0]`
  - long form: `tmp = a0, a0 = 1234, jump [tmp]`

Short form is only accepted if the two passed registers are not the same. For the case that they are, only the long form is accepted. The outputted disassembly also follows this rule. When writing assembly, the long form will always be accepted.

### DIFF
--- a/crates/polkavm-common/src/assembler.rs
+++ b/crates/polkavm-common/src/assembler.rs
@@ -377,15 +377,18 @@ pub fn assemble(code: &str) -> Result<Vec<u8>, String> {
                                 emit_and_continue!(MaybeInstruction::LoadImmAndJump(dst, value, label.to_owned()));
                             }
                             if let Some((base, offset)) = parse_indirect_memory_access(line) {
-                                if dst == base {
-                                    return Err(format!("cannot parse line {nth_line}: \"{original_line}\"; hint: use long form."));
-                                }
-                                emit_and_continue!(Instruction::load_imm_and_jump_indirect(
+                                let instruction = Instruction::load_imm_and_jump_indirect(
                                     dst.into(),
                                     base.into(),
                                     value as u32,
                                     offset as u32
-                                ));
+                                );
+                                
+                                if dst == base {
+                                    return Err(format!("cannot parse line {nth_line}, expected: \"{instruction}\""));
+                                }
+
+                                emit_and_continue!(instruction);
                             }
                         }
                     }
--- a/crates/polkavm-common/src/assembler.rs
+++ b/crates/polkavm-common/src/assembler.rs
@@ -35,17 +35,34 @@ fn parse_indirect_memory_access(text: &str) -> Option<(Reg, i32)> {
     }
 }
 
-/// Parses the memory access jump of load_imm_and_jump_indirect long form:
-/// `jump [tmp]` or `jump [tmp + {offset}]`
-fn parse_indirect_jump_memory_access(text: &str) -> Option<i32> {
-    let text = text.trim().strip_prefix("jump [")?.strip_suffix(']')?;
+/// Parses the long form of load_imm_and_jump_indirect:
+/// `tmp = {base}, {dst} = {value}, jump [tmp + {offset}]`, where `dest == base` is allowed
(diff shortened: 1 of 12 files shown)


crates/polkavm-common/src/assembler.rs
```diff
@@ -35,6 +35,20 @@ fn parse_indirect_memory_access(text: &str) -> Option<(Reg, i32)> {
...
+    if let Some(index) = text.find('+') {
+        let _ = text[..index].trim().strip_prefix("tmp")?;
+        let offset = parse_imm(&text[index + 1..])?;
+        Some(offset)
+    } else {
+        let _ = text.strip_prefix("tmp")?;
+        Some(0)
+    }
+}
+
```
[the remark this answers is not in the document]
subotic: > As far as I can see this will be parsed successfully by this: `jump [tmpXXX+1]`, `jump [tmp foobar 1234]`.
> 
Yes, you are right.  Need to make parsing of this bit more restrictive.

crates/polkavm-common/src/assembler.rs
```diff
@@ -336,6 +376,17 @@ pub fn assemble(code: &str) -> Result<Vec<u8>, String> {
...
+                            if let Some((base, offset)) = parse_indirect_memory_access(line) {
+                                if dst == base {
+                                    return Err(format!("cannot parse line {nth_line}: \"{original_line}\"; hint: use long form."));
+                                }
+                                emit_and_continue!(Instruction::load_imm_and_jump_indirect(
+                                    dst.into(),
+                                    base.into(),
+                                    value as u32,
+                                    offset as u32
+                                ));
```
koute, line 388: We could just print out what it expects in the error message; something like this maybe:

```suggestion
                                let instruction = Instruction::load_imm_and_jump_indirect(
                                    dst.into(),
                                    base.into(),
                                    value as u32,
                                    offset as u32
                                );
                                
                                if dst == base {
                                    return Err(format!("cannot parse line {nth_line}, expected: \"{instruction}\""));
                                }

                                emit_and_continue!(instruction);
```

crates/polkavm-common/src/assembler.rs
```diff
@@ -35,6 +35,37 @@ fn parse_indirect_memory_access(text: &str) -> Option<(Reg, i32)> {
     }
 }
 
+/// Parses the long form of load_imm_and_jump_indirect:
+/// `tmp = {base}, {dst} = {value}, jump [tmp + {offset}]`, where `dest == base` is allowed
+fn parse_load_imm_and_jump_indirect(line: &str) -> Option<(Reg, Reg, i32, i32)> {
```
koute, line 40:
```suggestion
fn parse_load_imm_and_jump_indirect_with_tmp(line: &str) -> Option<(Reg, Reg, i32, i32)> {
```

Since this can only handle the `tmp` form. Kinda long name, but I don't have any better ideas.

subotic: yes, it also crossed my mind, but wasn't sure about about it. My other idea was `parse_load_imm_and_jump_indirect_long_form` but this is even longer.

crates/polkavm-common/src/assembler.rs
```diff
@@ -35,6 +35,41 @@ fn parse_indirect_memory_access(text: &str) -> Option<(Reg, i32)> {
...
+    let base = parse_reg(line[..index].trim())?;
+    let line = line[index + 1..].trim();
+
+    let index = line.find('=')?;
+    let dst = parse_reg(line[..index].trim())?;
+    let line = line[index + 1..].trim();
+
+    let index = line.find(',')?;
+    let value = parse_imm(line[..index].trim())?;
+    let text = line[index + 1..].trim().strip_prefix("jump [")?.strip_suffix(']')?;
```
koute, line 57: There's still a hardcoded space here. (:

subotic: Sorry about that.  I was to eager to finish it. Will fix everything today.

PR #145, opened 2024-07-18
add tests to invalid dynamic jumps

alk0x1:
Add tests for two cases of dynamic jumps that should return a `ExecutionError::Trap`

### DIFF
--- a/crates/polkavm/src/tests.rs
+++ b/crates/polkavm/src/tests.rs
@@ -517,21 +517,37 @@ impl TestInstance {
     }
 }
 
-fn basic_test_blob_invalid_jump() -> ProgramBlob {
+fn blob_null_pointer_djump() -> ProgramBlob {
     let mut builder = ProgramBlobBuilder::new();
     builder.set_rw_data_size(0x4000);
     builder.add_export_by_basic_block(0, b"main");
     builder.set_code(
         &[
-            asm::jump_indirect(Reg::A0, 15),
+            asm::jump_indirect(Reg::A0, 0x0),
         ],
-        &[],
+        &[0x0],
     );
     ProgramBlob::parse(builder.into_vec().into()).unwrap()
 }
 
-fn test_invalid_jump_trap(config: Config) {
+fn blob_miss_aligned_pointer_djump() -> ProgramBlob {
+    let mut builder = ProgramBlobBuilder::new();
+    builder.set_rw_data_size(0x4000);
+    builder.add_export_by_basic_block(0, b"main");
+    builder.set_code(
+        &[
+            asm::load_imm(Reg::A0, 0x1000),
+            asm::jump_indirect(Reg::A0, 1)
+        ],
+        &[1],
+    );
+    
+    ProgramBlob::parse(builder.into_vec().into()).unwrap()
+}
+
+fn setup_instance(blob: ProgramBlob) -> crate::Instance<()>{
     let _ = env_logger::try_init();
+
     let mut config = crate::Config::default();
     config.set_backend(Some(crate::BackendKind::Compiler));
     config.set_sandbox(Some(crate::SandboxKind::Linux));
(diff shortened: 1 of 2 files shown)


crates/polkavm/src/tests.rs
```diff
@@ -517,6 +517,88 @@ impl TestInstance {
     }
 }
 
+fn blob_null_pointer_djump() -> ProgramBlob {
+    let mut builder = ProgramBlobBuilder::new();
+    builder.set_rw_data_size(0x4000);
```
koute, line 522: Not needed for these tests.

```suggestion
```

crates/polkavm/src/tests.rs
```diff
@@ -517,6 +517,88 @@ impl TestInstance {
...
+    builder.add_export_by_basic_block(0, b"main");
+    builder.set_code(
+        &[
+            asm::jump_indirect(Reg::A0, 0x0),
+        ],
+        &[0x0],
```
koute, line 528:
```suggestion
        &[],
```

First entry of the jump table is already implicitly zero.

crates/polkavm/src/tests.rs
```diff
@@ -517,6 +517,88 @@ impl TestInstance {
     }
 }
 
+fn blob_null_pointer_djump() -> ProgramBlob {
```
koute, line 520: No need to put these in separate functions; should be inlined in the test that use these.

crates/polkavm/src/tests.rs
```diff
@@ -517,6 +517,88 @@ impl TestInstance {
...
+
+    let module = Module::from_blob(&engine, &Default::default(), blob).unwrap();
+    let instance_pre = linker.instantiate_pre(&module).unwrap();
+
+    instance_pre.instantiate().unwrap()
+}
+
+#[test]
+#[ignore]
+fn test_invalid_jump_trap() {
```
koute, line 566: These should be three separate tests, instead of all in one, and it should be executed for all of the backends, like other tests.

crates/polkavm/src/tests.rs
```diff
@@ -517,6 +517,88 @@ impl TestInstance {
...
+}
+
+fn blob_miss_aligned_pointer_djump() -> ProgramBlob {
+    let mut builder = ProgramBlobBuilder::new();
+    builder.set_rw_data_size(0x4000);
+    builder.add_export_by_basic_block(0, b"main");
+    builder.set_code(
+        &[
+            asm::load_imm(Reg::A0, 0x1000),
+            asm::jump_indirect(Reg::A0, 1)
```
koute, line 540:
```suggestion
            asm::jump_indirect(Reg::A0, 1)
```

crates/polkavm/src/tests.rs
alk0x1: The problem is, if I put the test in the `run_tests!` macro I cannot use the `#[ignore]`

koute: You can just "ignore" them manually with an early `return` and a `// TODO` explaining why.

crates/polkavm/src/tests.rs
```diff
@@ -517,6 +517,46 @@ impl TestInstance {
...
+
+    let state_args = StateArgs::default();
+    let ext_main = instance.module().lookup_export("main").unwrap();
+
+    let result = instance.call(state_args, CallArgs::new(&mut (), ext_main));
+
+    match result {
+        Ok(()) => panic!("Expected ExecutionError::Trap, but got Ok"),
+        Err(e) => assert!(matches!(e, ExecutionError::Trap(_)), "Expected ExecutionError::Trap, but got {:?}", e),
+    }
```
jarkkojs, line 562: More idiomatic Rust:

```rust
if let Err(e) == instance.call(state_args, CallArgs::new(&mut (), ext_main)) {
    assert!(matches!(e, ExecutionError::Trap(_)), "Expected ExecutionError::Trap, but got {:?}", e);
} else {
    panic!("Expected ExecutionError::Trap, but got Ok");
}

```

PR #147, opened 2024-07-28
enable creating invalid programs

subotic:


crates/polkavm-common/src/writer.rs
```diff
@@ -34,6 +34,19 @@ impl From<(u32, Instruction)> for InstructionBuffer {
     }
 }
 
+impl From<Vec<u8>> for InstructionBuffer {
+    fn from(value: Vec<u8>) -> Self {
+        let mut bytes = [0u8; program::MAX_INSTRUCTION_LENGTH];
+        let slice_size = value.len().min(program::MAX_INSTRUCTION_LENGTH);
```
koute, line 40: Shouldn't silently truncate. Add a `TryFrom` impl instead.

crates/polkavm-common/src/writer.rs
```diff
@@ -120,7 +151,7 @@ impl ProgramBlobBuilder {
         self.exports.push((target_basic_block, ProgramSymbol::new(symbol.into())));
     }
 
-    pub fn set_code(&mut self, code: &[Instruction], jump_table: &[u32]) {
```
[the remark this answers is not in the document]
subotic: @koute Done. I have a question regarding this though. Sorry if it is a stupid one. As I understand it, the reason behind `&[Instruction]` was to be able to borrow a slice of `Instruction`'s. This has now changed into `&[impl Into<InstructionOrBytes> + Copy]`.  So far ok. Then now we have as the next line:
`let code: Vec<InstructionOrBytes> = code.iter().map(|inst| (*inst).into()).collect();`
Isn't this now making a copy of the contents of the slice and storing it inside a `Vec`? Couldn't we at this point redefine `code` from `&[impl Into<InstructionOrBytes> + Copy]` to `code: Vec<InstructionOrBytes>`, or what am I missing?

crates/polkavm-common/src/writer.rs
```diff
@@ -148,25 +181,40 @@ impl ProgramBlobBuilder {
...
-                target_nth_instruction
-            });
-
-            let entry = SerializedInstruction {
-                instruction,
-                bytes: InstructionBuffer::from((position, instruction)),
-                target_nth_instruction,
-                position,
-            };
+            if let InstructionEx::Instruction(mut instruction) = (*instruction).into() {
```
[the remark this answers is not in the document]
subotic: refactored

crates/polkavm-common/src/writer.rs
subotic: I removed this, as it is not needed anymore. `InstructionBuffer` is used directly inside `InstructionOrBytes`.

tools/spectool/spec/src/inst_load_imm_and_jump_nok___.txt
```diff
@@ -0,0 +1,5 @@
+pub @main:
+pub @expected_exit:
+    a0 = 1234, jump @invalid_offset 25
```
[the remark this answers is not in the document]
subotic: @koute I have some questions regarding this. Will pose them in the matrix chat.

crates/polkavm-common/src/writer.rs
```diff
@@ -137,8 +156,10 @@ impl ProgramBlobBuilder {
         basic_block_to_instruction_index.push(0);
 
         for (nth_instruction, instruction) in code.iter().enumerate() {
-            if instruction.opcode().starts_new_basic_block() {
-                basic_block_to_instruction_index.push(nth_instruction + 1);
+            if let InstructionOrBytes::Instruction(inst) = instruction {
+                if inst.opcode().starts_new_basic_block() {
+                    basic_block_to_instruction_index.push(nth_instruction + 1);
+                }
```
subotic, line 162: @koute This will pose a problem, if we ever would like to add a label to the instruction when writing a program in assembler, as `load_imm_and_jump` is actually starting a new block but will not be taken into consideration when it is provided as `bytes`.

crates/polkavm-common/src/assembler.rs
```diff
@@ -686,36 +694,39 @@ pub fn assemble(code: &str) -> Result<Vec<u8>, String> {
...
+                TargetKind::Label(label) => {
+                    let Some(&target_index) = label_to_index.get(&*label) else {
+                        return Err(format!("label is not defined: \"{label}\""));
+                    };
+                    code.push(Instruction::load_imm_and_jump(dst.into(), value as u32, target_index).into());
+                }
+                TargetKind::Offset(offset) => {
+                    // TODO: unclear how to deal with negative offsets
+                    let instruction = Instruction::load_imm_and_jump(dst.into(), value as u32, offset as u32);
+                    code.push(InstructionBuffer::from((0, instruction)).into());
```
subotic, line 722: @koute Will this actually work for negative offsets and if so, how?

subotic: So this one I could answer for myself in the meantime. The negative value is preserved because the `u32` is reinterpreted as an `i32` for offsets, when deserializing.

crates/polkavm-common/src/writer.rs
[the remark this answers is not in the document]
koute: Yes. This is fine.

> Couldn't we at this point redefine `code` from `&[impl Into<InstructionOrBytes> + Copy]` to `code: Vec<InstructionOrBytes>`, or what am I missing?

We could, but that will require changing all of the places this function is used; this way it's going to be backwards compatible.

crates/polkavm-common/src/writer.rs
koute: Hm, okay, good point. In that case maybe we could add a flag to the raw bytes to specify whether it's supposed to start a new basic block?

crates/polkavm-common/src/assembler.rs
```diff
@@ -379,7 +385,9 @@ pub fn assemble(code: &str) -> Result<Vec<u8>, String> {
                     if let Some(value) = parse_imm(&rhs[..index]) {
                         if let Some(line) = rhs[index + 1..].trim().strip_prefix("jump") {
                             if let Some(label) = line.trim().strip_prefix('@') {
-                                emit_and_continue!(MaybeInstruction::LoadImmAndJump(dst, value, label.to_owned()));
+                                emit_and_continue!(MaybeInstruction::LoadImmAndJump(dst, value, TargetKind::Label(label.to_owned())));
+                            } else if let Some(offset) = parse_imm(line) {
+                                emit_and_continue!(MaybeInstruction::LoadImmAndJump(dst, value, TargetKind::Offset(offset)));
```
koute, line 390: This should enforce that there's either a `+` or a `-` there at the start. (To make it explicit that this is supposed to be a *relative* offset, and not an absolute one like if a label is used.)

subotic: Ah, got it now. Will make the parsing more strict.

crates/polkavm-common/src/writer.rs
subotic: Hm,  yes. I will try it out and report back.

crates/polkavm-common/src/writer.rs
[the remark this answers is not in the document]
subotic: Backwards compatibility. Got it. Thanks for the explanation.

PR #149, opened 2024-08-07
64bit support: front end

xermicus:
The first PR for riscv64 support implementing it the front end. Should leave the 32bit instruction set completely intact.

Changes:
- Be generic over ELF header bitness
- Few new relocations
- Instruction decoding for RV64IMCA
- Added many new `D` and `W` instructions that need to be implement in the next PR


crates/polkavm/src/compiler.rs
```diff
@@ -198,12 +217,21 @@ impl<'a, S> CompilerVisitor<'a, S> where S: Sandbox {
...
+    where
+        S: Sandbox,
+    {
         // Finish with a trap in case the code doesn't end with a basic block terminator.
         ArchVisitor(&mut self).trap();
 
         let epilogue_start = self.asm.len();
-        self.code_offset_to_native_code_offset.push((self.code_length, epilogue_start as u32));
+        self.code_offset_to_native_code_offset
+            .push((self.code_length, epilogue_start as u32));
```
koute, line 234: This is reformatted for some reason? Can you revert the reformatting changes? (Otherwise it unnecessarily complicates resolving merge conflicts later.)

crates/polkavm-common/src/program.rs
```diff
@@ -1383,23 +1405,38 @@ define_opcodes! {
     // Instructions with args: reg, reg, reg
     [
         add                                      = 8,
+        addw                                     = 101,
```
koute, line 1408: Hmm... so we can do this in two ways:

1. Redefine existing instructions to be 64-bit, and define a new 32-bit instructions.
2. Keep the current instructions as 32-bit, and define new one 64-bit ones.

I'm thinking maybe (2) would be better? Then the current implementation of what we have could stay as-is and be shared with both 32-bit and 64-bit. Thoughts?

crates/polkavm/src/interpreter.rs
```diff
@@ -905,6 +905,14 @@ impl<'a, 'b, const DEBUG: bool> InstructionVisitor for Visitor<'a, 'b, DEBUG> {
         self.set3(d, s1, s2, |s1, s2| u32::from((s1 as i32) < (s2 as i32)))
     }
 
+    fn set_less_than_unsigned_w(&mut self, _d: RawReg, _s1: RawReg, _s2: RawReg) -> Self::ReturnTy {
+        todo!()
```
koute, line 909: Make all of these `todo`s into traps until we get full support implemented.

crates/polkavm/src/compiler.rs
xermicus: I'm not sure maybe new rustfmt version? I can revert those and any other formatting changes I spot

koute: There's a rust-toolchain.toml in the repo though, so it shouldn't have used a new rustfmt version. Unless you nuked it on your copy?

crates/polkavm-common/src/program.rs
xermicus: I think both would work. I opted for approach 1. because that maps the actual RISC-V instructions to those. I think they could still be shared, as on 32bit we'd just wire them to the W versions. I thought if we expect people to default on 64bit in the future or even strip 32bit support, it might then seem a bit weird. However you are right in defining new `D` instructions instead of the `W` would simplify the implementation for 64bit.

crates/polkavm/src/compiler.rs
xermicus: Hmm, right, no I didn't nuke that, neither didn't change it manually :sweat_smile:

crates/polkavm-common/src/program.rs
koute: Well, considering our instruction set is pretty much independent of RISC-V at this point I think not having a direct mapping is fine, and we can always easily change things later with a quick search & replace, so I think making the implementation as easy as possible for now would be a good idea?

So for now could we make the 64-bit instructions the new ones, and as far as naming them just name them as `${name}64` so that it's clear they're 64 bit? (e.g. `add64`) If we decide to get rid of 32-bit support in the future it should be very easy to rename them.

crates/polkavm/src/compiler.rs
xermicus: I think I solved this mystery #152

crates/polkavm/src/interpreter.rs
xermicus: Done

crates/polkavm-common/src/program.rs
xermicus: Changed the new instructions to 64bit. This also has the advantage that they are disambiguated at the linking stage already.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3130,25 +3349,41 @@ impl OperationKind {
...
         #[allow(clippy::unnecessary_cast)]
         match self {
-            Self::Add => lhs.wrapping_add(rhs),
-            Self::Sub => lhs.wrapping_sub(rhs),
-            Self::And => lhs & rhs,
-            Self::Or => lhs | rhs,
-            Self::Xor => lhs ^ rhs,
-            Self::SetLessThanUnsigned => i32::from((lhs as u32) < (rhs as u32)),
-            Self::SetLessThanSigned => i32::from((lhs as i32) < (rhs as i32)),
+            // TODO
```
koute, line 3352: So for these I think we'd just change the arguments and the return value from `i32`s to `i64`s, and for 32-bit operations apply appropriate masking?

One complication with 64-bit constant evaluation is that if the result is supposed to end up as part of an `_imm` instruction then the result must fit into 32-bit, since we in general can't go more than 32-bit for immediates. (Since we want to limit how long our instructions can get; no instruction should need more than 16 bytes in total to decode.)

xermicus: Sounds sensible. Do you want to have this in this PR?

koute: Hm, up to you. Currently as it stands this makes the whole thing totally broken on 64-bit, but since 64-bit wouldn't work anyway I don't particularly care, except I guess in that case the bodies of these should probably be replaced with `todo!` to make sure we don't accidentally forget to fix it.

xermicus: Yeah its not the only thing missing. I agree those should better be `todo!()`, changed it.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -1295,7 +1313,7 @@ fn parse_extern_metadata_impl(
 
     let RelocationKind::Abs {
         target: symbol_location,
-        size: RelocationSize::U32,
+        size: RelocationSize::U32 | RelocationSize::U64,
```
koute, line 1316: The `read_u32` before this should be changed too, and we should enforce the relocation size depending on the bitness.

```rust
if elf.is_64() {
    b.read_u64()?;
} else {
    b.read_u32()?;
};

let symbol_location = match symbol_relocation {
    RelocationKind::Abs { target, size: RelocationSize::U64 } if elf.is_64() => target,
    RelocationKind::Abs { target, size: RelocationSize::U32 } if !elf.is_64() => target,
    _ => return Err(...);
}
```

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -1768,7 +1830,29 @@ fn convert_instruction(
...
+            };
+
+            let Some(src) = cast_reg_non_zero(src)? else {
+                return Err(ProgramFromElfError::other(
+                    "found an atomic load with a zero register as the source",
+                ));
+            };
+
+            emit(InstExt::Basic(BasicInst::LoadIndirect {
+                kind: LoadKind::I32,
```
koute, line 1855: LoadReserved64 is `LoadKind::I32` and `LoadReserved` is either U64 or U32? This looks wrong, naming-wise. `LoadReserved` is supposed to be `LR` and `LoadReserved64` is supposed to be `LR.W`, correct?

When I said that we should have 64-bit instructions be suffixed with 64 I only meant PolkaVM instructions. (:

So, for example, PolkaVM's `add` (or `add32`, to which we might rename `add` to in the future when 64-bit becomes the "default") should always add two `i32`s, and PolkaVM's `add64` should always add two `u64`s. Straightforward naming, they do what it says on the tin. But here the `Inst` is a RISC-V instruction, so the same naming scheme is wrong and doesn't make any sense. The `64` suffix suggests to me that something operates on 64-bit values, but here it operates on 32-bits.

So e.g. in this case `LoadReserved64` should be renamed to be either `LoadReservedW` (since it corresponds to RISC-V's `LR.W`) or `LoadReservedI32` (because it's an `LR` which loads an `i32`), whichever you prefer. (The `crate::riscv::Inst` is supposed to closely map RISC-V, just with nicer naming)

crates/polkavm-common/src/program.rs
```diff
@@ -1660,6 +1703,41 @@ impl<'a, 'b> InstructionVisitor for InstructionFormatter<'a, 'b> {
         write!(self, "{d} = {s1} <s {s2}")
     }
 
+    fn set_less_than_unsigned_64(&mut self, d: RawReg, s1: RawReg, s2: RawReg) -> Self::ReturnTy {
+        let d = self.format_reg(d);
+        let s1 = self.format_reg(s1);
+        let s2 = self.format_reg(s2);
+        write!(self, "i64 {d} = {s1} <u {s2}")
```
koute, line 1710: Hmm... I'm not sure putting `i64` at the start is the way to go here?

This is especially confusing when you have an instruction "set_less_than_**unsigned**_64" (emphasis on "unsigned") and yet there is an `i64` at the start (which would suggest that it's signed)? (:

Maybe `{d} = {s1} <u64 {s2}` would be better?

Same with bitshifts, for example: `i64 {d} = i64 {s1} << i64 {s2}` - here the `i` is confusing as that would suggest that signedness matters (i.e. that it's an arithmetic shift), but it doesn't, and it's also duplicated three times which feels unnecessary (remember that we also have to parse this disassembly with our assembler, and in theory it should also be convenient to write, although I don't expect anyone to code much in PolkaVM assembly directly :P).

What do you think?

(In general the syntax for the disassembly is not final and open for discussion, and I welcome any potential proposals for improvement.)

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3669,10 +3893,18 @@ fn perform_constant_propagation(
                 let section = elf.section_by_index(target.section_index);
                 if section.is_allocated() && !section.is_writable() {
                     let value = match kind {
+                        LoadKind::U64 => section
+                            .data()
+                            .get(target.offset as usize..target.offset as usize + 8)
+                            .map(|xs| u64::from_le_bytes([xs[0], xs[1], xs[2], xs[3], xs[4], xs[5], xs[6], xs[7]]) as i32),
```
koute, line 3899: The cast's incorrect here. It's fine to have code here which is going to extract the `u64`, but we shouldn't cast it to `i32` - this should be a `todo!`.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -6449,7 +6776,7 @@ fn harvest_code_relocations(
 
             match lo_inst {
                 Inst::Load {
-                    kind: LoadKind::U32,
+                    kind: LoadKind::U32 | LoadKind::U64 | LoadKind::I32,
```
koute, line 6779: Should be strict according to whether the ELF is 64-bit or not, instead of accepting any kind on both bitnesses.

Also, do we need all three? Wouldn't the `LoadKind::U64` suffice for 64-bit?

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -6612,15 +6942,27 @@ impl Config {
 }
 
 pub fn program_from_elf(config: Config, data: &[u8]) -> Result<Vec<u8>, ProgramFromElfError> {
-    let mut elf = Elf::parse(data)?;
+    match Elf::<object::elf::FileHeader32<object::endian::LittleEndian>>::parse(data) {
+        Ok(elf) => program_from_elf_internal(config, elf),
+        Err(ProgramFromElfError(ProgramFromElfErrorKind::FailedToParseElf(e))) if e.to_string() == "Unsupported ELF header" => {
+            let _elf = Elf::<object::elf::FileHeader64<object::endian::LittleEndian>>::parse(data)?;
+            todo!("64bit isn't fully supported yet");
+            //program_from_elf_internal(config, _elf)
```
koute, line 6950:
```suggestion
```

crates/polkavm-linker/src/riscv.rs
```diff
@@ -1044,6 +1357,9 @@ impl Inst {
                     | (u32::from(acquire) << 26)
                     | (0b00011 << 27),
             ),
+            Inst::StoreConditional64 { .. } | Inst::LoadReserved64 { .. } => {
+                unreachable!("internal error: encoding a 64bit instruction in a 32bit context")
```
koute, line 1361: Should return a `None`.

crates/polkavm/src/compiler/amd64.rs
```diff
@@ -197,6 +197,121 @@ where
 {
     pub const PADDING_BYTE: u8 = 0x90; // NOP
 
+    pub fn _load_i32(&mut self, _: RawReg, _: u32) {
+        todo!()
+    }
```
koute, line 202: These should either be:

a) deleted (since right now it's completely dead code, and the names aren't even updated to the new naming scheme) and the `todo!`s in functions in `compiler.rs` implemented to visit a trap,
b) renamed (some of these still have old names), the functions in `compiler.rs` implemented to call these, and the `todo!`s changed to trap (like for the interpreter).

In general the new work-in-progress 64-bit implementation should have no visible effects on the current 32-bit support (as implemented right now using the new 64-bit opcodes will trigger a `todo!` panic even in a 32-bit code).

crates/polkavm-linker/src/program_from_elf.rs
xermicus: Good catch thanks!

crates/polkavm-linker/src/riscv.rs
xermicus: Agreed, I changed it to return an error instead (as `convert_instruction` doesn't return `Option`s).

crates/polkavm-linker/src/program_from_elf.rs
xermicus: Changed

crates/polkavm-common/src/program.rs
xermicus: I agree that isn't ideal. I fixed them to more closely resemble the 32bit variants

koute:
Thanks!

PR #155, opened 2024-08-17
Always open logging pipe with child

xermicus:
On `Linux 6.10.5-arch1-1` the hello world example with uffd enabled only works with logs enabled (without the logging pipe set up, the child traps during spawn if dynamic paging is enabled).

I noticed that the zygote does write all logs regardless of the log level, so it should be fine to have the pipe even without logging since the sandbox  filters them according to log level anyways.

This PR together with #154 fixes the hello world example with dynamic paging.


xermicus:
Fixed in `0.11.0`

PR #156, opened 2024-08-18
Host function Stubs/Test Vectors v0.1

sourabhniyogi:
## Host function Test Vectors

This contributes test vectors for JAM implementers to implement host functions of Appendix B of GP, all of which read + write registers just like Appendix A of GP but further interact with DA + PMT storage, setup and take down PVMs, and so on.  It is necessary to implement host functions to build elementary meaningful Work Packages / JAM Services (refine, accumulate, transfer, auth PVM)

This is NOT an official test vector set (and really just a starting point since there is no PASS / FAIL condition), but a community contribution from a JAM implementer for people to start implementing Appendix B host functions.  In particular, **Gav cautions**: "to use opcodes which are not yet in use in place of host functions I would caution against it. Opcodes which are not in use should panic as an illegal instruction. If they do anything else then it will fail test vectors."

This directory has  host function tests for JAM implementers, covering the 23 host functions described in Appendix B:
 * `read`  = 130 (0x82)
 * `write` = 131 (0x83)
 * `solicit` = 139 (0x8b)
 * `forget` = 140 (0x8c)
 * `import` = 147 (0x93)
 * `export` = 148 (0x94)
 * `historical_lookup` = 141 (0x8d)
 * `lookup` = 129 (0x81)
 * `assign` = 149 (0x94)
 * `checkpoint` = 135 (0x87)
 * `designate` = 134 (0x86)
 * `egas` = 128 (0x80)
 * `empower` = 133 (0x85)
 * `expunge` = 148 (0x94)
 * `info` = 132 (0x84)
 * `invoke` = 147 (0x93)
 * `machine` = 144 (0x90)
 * `new` = 150 (0x96)
 * `peek` = 145  (0x91)
 * `poke` = 146 (0x92)
 * `quit` = 138 (0x8a)
 * `transfer` = 137 (0x89)
 * `upgrade` = 136 (0x88)

Each of these host functions are like `fallthrough` in not taking any arguments and result in just one byte in assembled byte code.  To avoid collisions with opcodes in Appendix A, 128 is added.  23 Opcodes (128-150) are used up with this method.

For each of the 23 host functions, there is at least one assembly file using the host function in  `tools/spectool/spec/src/hostfunctions/`  See README.md for the assembly code and the directory for "disassemble". 

Using this byte code, JAM implementers can implement host functions.


sourabhniyogi:
@koute There is no need for this as "ecalli" is the bridge between Appendix A and Appendix B.

PR #168, opened 2024-09-17
add post checks to spectool

subotic:
Add post checks for gas and regs.


subotic:
@koute Something like this? I've added some pre and post directives to `inst_fallthrough` which should fail for one `reg` and `gas` (just as an example to test if it works).

koute (approved):

koute:
Thanks!

PR #172, opened 2024-09-19
Add post checks for existing tests in tools/spectool

### LINKED ISSUE #170: Add post checks for existing tests in `tools/spectool`
https://github.com/koute/polkavm/pull/168

subotic:
resolves #170

### DIFF
--- a/tools/spectool/spec/src/inst_store_imm_indirect_u16_with_offset_nok.txt
+++ b/tools/spectool/spec/src/inst_store_imm_indirect_u16_with_offset_nok.txt
@@ -1,7 +1,10 @@
 pre: a0 = 0x20FFE
+pre: gas = 10000
 
 %rw_data_size = 2
 
 pub @main:
 pub @expected_exit:
     u16 [a0 + 1] = 0x1234
+
+post: gas = 9998
--- a/tools/spectool/spec/src/inst_add.txt
+++ b/tools/spectool/spec/src/inst_add.txt
@@ -1,5 +1,9 @@
 pre: a0 = 1
 pre: a1 = 2
+pre: gas = 10000
 
 pub @main:
     a2 = a0 + a1
+
+post: a2 = 3
+post: gas = 9998
--- a/tools/spectool/spec/src/inst_add_imm.txt
+++ b/tools/spectool/spec/src/inst_add_imm.txt
@@ -1,4 +1,9 @@
 pre: a0 = 1
+pre: gas = 10000
+
 
 pub @main:
     a2 = a0 + 2
+
+post: a2 = 3
+post: gas = 9998
--- a/tools/spectool/spec/src/inst_add_with_overflow.txt
+++ b/tools/spectool/spec/src/inst_add_with_overflow.txt
@@ -1,5 +1,9 @@
 pre: a0 = -1
 pre: a1 = 2
+pre: gas = 10000
 
 pub @main:
     a2 = a0 + a1
+
+post: a2 = 1
+post: gas = 9998
--- a/tools/spectool/spec/src/inst_and.txt
+++ b/tools/spectool/spec/src/inst_and.txt
@@ -1,5 +1,9 @@
 pre: a0 = 0b101
 pre: a1 = 0b011
+pre: gas = 10000
 
 pub @main:
     a2 = a0 & a1
+
+post: a2 = 0b001
+post: gas = 9998
--- a/tools/spectool/spec/src/inst_and_imm.txt
+++ b/tools/spectool/spec/src/inst_and_imm.txt
@@ -1,4 +1,8 @@
 pre: a0 = 0b101
+pre: gas = 10000
 
 pub @main:
     a2 = a0 & 0b011
+
+post: a2 = 0b001
+post: gas = 9998
--- a/tools/spectool/spec/src/inst_branch_eq_imm_nok.txt
+++ b/tools/spectool/spec/src/inst_branch_eq_imm_nok.txt
@@ -1,3 +1,5 @@
+pre: gas = 10000
+
 pub @main:
     a0 = 1234
     jump @target if a0 == 1235
@@ -5,3 +7,6 @@ pub @expected_exit:
     trap
 @target:
     a0 = 0xdeadbeef
+
+post: a0 = 1234
+post: gas = 9997
--- a/tools/spectool/spec/src/inst_branch_eq_imm_ok.txt
+++ b/tools/spectool/spec/src/inst_branch_eq_imm_ok.txt
@@ -1,6 +1,11 @@
+pre: gas = 10000
+
 pub @main:
     a0 = 1234
     jump @target if a0 == 1234
     trap
 @target:
     a0 = 0xdeadbeef
+
+post: a0 = 0xdeadbeef
+post: gas = 9996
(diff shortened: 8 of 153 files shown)


koute:
So for those post checks which are now failing due to the behavior being wrong, can you make those into `// TODO: post: ...failing condition here...` comments so that we can merge this?

Also, some of the values you've put in are incorrect. (:

For example:

> inst_load_u8_trap.txt: unexpected gas: 9998 (expected: 9999)

```
pre: a0 = 0
pre: gas = 10000

pub @main:
pub @expected_exit:
    a0 = u8 [0x20000]

post: a0 = 0
post: gas = 9999
```

The correct final gas here *is* 9998, so the `post` condition is wrong!

This might not be in the GP yet (because the gas cost model is not done yet), but remember that we currently don't charge gas on each instruction (for efficiency reasons). We charge gas when we enter a given basic block for the whole basic block! So here we have two instructions: a load from memory, and an implicit trap. So this gives 2 gas consumed. Sure, in this case the load will interrupt the execution because it accesses memory that's inaccessible, but gas-wise that doesn't matter.

(Of course, there's a possibility that this could be changed, but for now that's how it works.)

subotic:
> So for those post checks which are now failing due to the behavior being wrong, can you make those into `// TODO: post: ...failing condition here...` comments so that we can merge this?

Sure, will do.

> Also, some of the values you've put in are incorrect. (:
>
> This might not be in the GP yet (because the gas cost model is not done yet), but remember that we currently don't charge gas on each instruction (for efficiency reasons). We charge gas when we enter a given basic block for the whole basic block! So here we have two instructions: a load from memory, and an implicit trap. So this gives 2 gas consumed. Sure, in this case the load will interrupt the execution because it accesses memory that's inaccessible, but gas-wise that doesn't matter.
> 
> (Of course, there's a possibility that this could be changed, but for now that's how it works.)

Oh, I wasn't aware of the "gas metering per basic block bit". Now the spent gas numbers make more sense. Thank your for explaining that. 

Maybe it would make sense to add this to the GP? My Scala PVM implementation does currently per instruction gas metering and would require refactoring, that will have a broader impact. I assume that it is similar for other teams.

koute:
> Maybe it would make sense to add this to the GP? My Scala PVM implementation does currently per instruction gas metering and would require refactoring, that will have a broader impact. I assume that it is similar for other teams.

Yes, but probably once we actually have a close-to-final gas cost model? You'll have to make changes for the final gas cost model anyway, so it's either we update the GP now and will have to update it later again (so two updates), or we'll only update it once later (so only one update).

subotic:
> > Maybe it would make sense to add this to the GP? My Scala PVM implementation does currently per instruction gas metering and would require refactoring, that will have a broader impact. I assume that it is similar for other teams.
> 
> Yes, but probably once we actually have a close-to-final gas cost model? You'll have to make changes for the final gas cost model anyway, so it's either we update the GP now and will have to update it later again (so two updates), or we'll only update it once later (so only one update).

I suppose, It depends on what is included in the final gas cost model. If only the cost per instruction values change, than that is not a big deal. If we are talking about changing gas metering from "per instruction" to "per basic block", then it is a bigger change.

I can only speak for my implementation, where if gas metering per basic block is necessary, then I would need to do quite some refactoring. In contrast, changing the per instruction gas cost values, this would be a trivial change.

On the other hand, adding basic block metering to the GP now and then removing it later, is also not a good option. So  I assume, that if things are that much unknown and in flux, to just leave it as it currently is, would be the better option.

koute:
> I can only speak for my implementation, where if gas metering per basic block is necessary, then I would need to do quite some refactoring.

Your PVM implementation is an interpreter, right? In that case you would have to completely rewrite it *anyway* because I don't expect anyone to be able to hit even the half-speed milestone on an interpreter. (:

If you [look at the benchmarks](https://github.com/koute/polkavm/blob/master/BENCHMARKS.md) (although these are slightly outdated as our interpreter should be faster now, maybe even twice as fast, I don't remember exactly) you can see that the performance difference between the recompiler and the interpreter is pretty big. Assuming the performance of my recompiler will be the "full speed" milestone bar (but most likely slightly lower to still allow for implementations which are not optimized pedal-to-the-metal to qualify), then "half speed" would suggest that you could run half the speed/take twice as much time as the recompiler. And I don't see even a very well optimized interpreter written in a systems language hitting that, never mind an interpreter which is written in a higher level language! (But maybe "half speed" will get redefined as some slightly lower threshold and not literally half; I don't know.)

That said, unless you're an expert in writing recompilers it doesn't make sense to start with a recompiler right off the bat; it's harder to write, harder to debug, and harder to modify. So it still makes sense to start with an interpreter (if nothing else you'll get some experience which will help you write a recompiler later), just don't expect to be able to use it right until the very end. (:

subotic:
> > I can only speak for my implementation, where if gas metering per basic block is necessary, then I would need to do quite some refactoring.
> 
> Your PVM implementation is an interpreter, right? In that case you would have to completely rewrite it _anyway_ because I don't expect anyone to be able to hit even the half-speed milestone on an interpreter. (:
> 
> If you [look at the benchmarks](https://github.com/koute/polkavm/blob/master/BENCHMARKS.md) (although these are slightly outdated as our interpreter should be faster now, maybe even twice as fast, I don't remember exactly) you can see that the performance difference between the recompiler and the interpreter is pretty big. Assuming the performance of my recompiler will be the "full speed" milestone bar (but most likely slightly lower to still allow for implementations which are not optimized pedal-to-the-metal to qualify), then "half speed" would suggest that you could run half the speed/take twice as much time as the recompiler. And I don't see even a very well optimized interpreter written in a systems language hitting that, never mind an interpreter which is written in a higher level language! (But maybe "half speed" will get redefined as some slightly lower threshold and not literally half; I don't know.)
> 
> That said, unless you're an expert in writing recompilers it doesn't make sense to start with a recompiler right off the bat; it's harder to write, harder to debug, and harder to modify. So it still makes sense to start with an interpreter (if nothing else you'll get some experience which will help you write a recompiler later), just don't expect to be able to use it right until the very end. (:

Yes, only an interpreter at the moment. As you stated, the hope is to first learn what needs to happen inside a VM, as this is my first time implementing something like this. So basically, I'm currently only going for correctness and not for speed. By the time I need to write a recompiler, I should have a much better understanding of what needs to be done. Furthermore, a reason why I'm trying to help out here a bit is to learn, where even with this issue, I'm learning something new. So thank you for your time. :)

tools/spectool/spec/src/inst_store_imm_indirect_u16_with_offset_nok.txt
subotic, line 1: @koute A general question regarding writing to memory. In the GP I see this round arrow on all memory write instructions:
image

I'm understanding this as "wrapping around" at the edge of the memory space. Should this program then panic during writing to memory?

subotic:
@koute can you please take a look again? issue should be good to go.

tools/spectool/spec/src/inst_store_imm_indirect_u16_with_offset_nok.txt
koute: The size of the address space is 32-bits, and the first 64kB are guaranteed to be always inaccessible (this is currently not in the GP yet as far as I remember, but it is a hard requirement as modern OSes like Linux don't even allow you to map any memory there for security reasons, and we don't want to force people to disable that security measure just to be able to run a hardware accelerated PVM). So yes, any access which wraps around the address space will, by definition, always panic.

koute (approved):
Stellar work, thank you!

PR #177, opened 2024-09-25
polkavm-linker: Modify RegValue so it can hold 64bit instructions

aman4150:
We are refactoring code so it can work with 64 bit ISA. The objective is to replace intermediate representation within the linker big enough for both 32 bit and 64 bit ISA.

### DIFF
--- a/crates/polkavm-linker/src/program_from_elf.rs
+++ b/crates/polkavm-linker/src/program_from_elf.rs
@@ -392,6 +392,10 @@ impl From<SectionTarget> for SectionIndex {
     }
 }
 
+fn i32_to_i64(x: i32) -> i64 {
+    i64::from(x)
+}
+
 fn extract_delimited<'a>(str: &mut &'a str, prefix: &str, suffix: &str) -> Option<(&'a str, &'a str)> {
     let original = *str;
     let start_of_prefix = str.find(prefix)?;
@@ -479,6 +483,14 @@ impl SectionTarget {
         SectionTarget {
             section_index: self.section_index,
             offset: u64::from(cb(offset as i32) as u32),
+        }
+    }
+
+    fn map_offset_i64(self, cb: impl FnOnce(i64) -> i64) -> Self {
+        let offset = self.offset as i64;
+        SectionTarget {
+            section_index: self.section_index,
+            offset: cb(offset) as u64,
         }
     }
 }
@@ -1796,7 +1808,10 @@ fn convert_instruction(
                 // The optimizer can take care of this later, but doing it early here is more efficient.
                 emit(InstExt::Basic(BasicInst::LoadImmediate {
                     dst,
-                    imm: OperationKind::from(kind).apply_const(0, imm),
+                    imm: OperationKind::from(kind)
+                        .apply_const(0, i32_to_i64(imm))
+                        .try_into()
+                        .expect("load immediate overflow"),
                 }));
             } else {
                 emit(InstExt::Basic(BasicInst::AnyAny {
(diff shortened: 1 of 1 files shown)


crates/polkavm-common/src/operation.rs
```diff
@@ -40,36 +69,78 @@ pub const fn rem(lhs: i32, rhs: i32) -> i32 {
...
+}
+
 #[inline]
 pub const fn mulh(lhs: i32, rhs: i32) -> i32 {
     ((lhs as i64).wrapping_mul(rhs as i64) >> 32) as i32
 }
 
+#[inline]
+pub const fn mulh64(lhs: i64, rhs: i64) -> i64 {
+    lhs.wrapping_mul(rhs >> 32)
```
koute, line 90: This is incorrect; the operands need to be cast to `i128` here, and the upper 64-bits should be returned. Same for the other helpers.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -6399,7 +6399,7 @@ fn emit_code(
                             if is_optimized {
                                 unreachable!("internal error: instruction with only constant operands: {op:?}")
                             } else {
-                                I::load_imm(dst, OperationKind::from(kind).apply_const(src1 as i32, src2 as i32) as u32)
+                                I::load_imm(dst, OperationKind::from(kind).apply_const(i64::from(src1), i64::from(src2)) as u32)
```
koute, line 6402: Do not use truncating casts that silently convert the types. If we have a codegen bug (as in - if we have an immediate argument that needs more than 32-bits) we want the code to panic if it cannot fit, instead of silently generating wrong code.

crates/polkavm-common/src/operation.rs
aman4150: Err. yes, will fix it.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: Hm, well, here `src1` and `src2` both are 32bit, so conversion would always happen.

But I think, you meant the intermediate representation should be 64bit...

aman4150: Operands are coming from `RegImm`, which we are not suppose to change for the while we have 32 bit support intact.

crates/polkavm-common/src/operation.rs
```diff
@@ -40,36 +69,78 @@ pub const fn rem(lhs: i32, rhs: i32) -> i32 {
...
+}
+
 #[inline]
 pub const fn mulhsu(lhs: i32, rhs: u32) -> i32 {
     ((lhs as i64).wrapping_mul(rhs as i64) >> 32) as i32
 }
 
+#[inline]
+pub const fn mulhsu64(lhs: i64, rhs: u64) -> i64 {
+    lhs.wrapping_mul(rhs as i64) >> 32
```
koute, line 100: Ahem, you fixed one, but this one is still broken :P

crates/polkavm-common/src/operation.rs
```diff
@@ -40,36 +69,78 @@ pub const fn rem(lhs: i32, rhs: i32) -> i32 {
...
+}
+
 #[inline]
 pub const fn mulhu(lhs: u32, rhs: u32) -> u32 {
     ((lhs as i64).wrapping_mul(rhs as i64) >> 32) as u32
 }
 
+#[inline]
+pub const fn mulhu64(lhs: u64, rhs: u64) -> u64 {
+    ((lhs as i128).wrapping_mul(rhs as i128) >> 32) as u64
```
koute, line 110: And this one's still broken too. This should be an unsigned multiplication.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3590,15 +3590,15 @@ impl RegValue {
                 dst,
                 target: AnyTarget::Data(target),
             }),
-            RegValue::Constant(imm) => Some(BasicInst::LoadImmediate { dst, imm }),
+            RegValue::Constant(imm) => Some(BasicInst::LoadImmediate { dst, imm: (imm as i32) }),
```
koute, line 3593: A truncating cast.

crates/polkavm-linker/src/program_from_elf.rs
koute: > Operands are coming from `RegImm`, which we are not suppose to change for the while we have 32 bit support intact.

No, they are not supposed to *ever* change, because even if we go full 64-bit only we won't be able to encode 64-bit immediates for every instruction. (:

Anyway, it doesn't matter where they are coming from. If there is *ever* a value that needs more than 32-bits here then this code is *totally broken*. This might not be the case right now, but this file is already a big ball of complex code and mud. I don't want to have to manually make sure this invariant is true and keep it in mind every time I change anything. So we want what is essentially an assertion that *if* by accident we get a too big of a value this code should panic and let us know that it's broken. That's why I said this should not be an infallible, truncating cast.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3470,25 +3470,25 @@ impl OperationKind {
...
+            Self::MulUpperUnsignedUnsigned => mulhu64(lhs as u64, rhs as u64) as i64,
+            Self::Div => div64(lhs as i64, rhs as i64),
+            Self::DivUnsigned => divu64(lhs as u64, rhs as u64) as i64,
+            Self::Rem => rem64(lhs as i64, rhs as i64),
+            Self::RemUnsigned => remu64(lhs as u64, rhs as u64) as i64,
+
+            Self::Eq => i64::from(lhs == rhs),
+            Self::NotEq => i64::from(lhs != rhs),
+            Self::SetGreaterOrEqualUnsigned => i64::from((lhs as u64) >= (rhs as u64)),
+            Self::SetGreaterOrEqualSigned => i64::from((lhs as i64) >= (rhs as i64)),
```
koute, line 3491: At least some of these, if not most of them, are incorrect for 32-bit.

Not every operation has the property that '(x op y) as u32 == (x as u32) op (y as u32)`.

For example `0xffffffff_u32 >> 32` and `0xffffffff_u64 >> 32` give an entirely different result, division is also broken, etc.

aman4150:
Thanks for the pointers @koute.

I've updated the PR. For this patch-set, I have not extended linker logic beyond 32 bit instructions, and just laid the ground work for 64 bit. In a follow up PR, I'll add support for 64bit instructions along with a bunch of tests to check linker 64 bit support.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3461,34 +3472,98 @@ impl From<BranchKind> for OperationKind {
...
+            }
+            Self::Sub => {
+                let lhs: i32 = lhs.try_into().expect("sub operand overflow!");
+                let rhs: i32 = rhs.try_into().expect("sub operand overflow!");
+                i64::from(lhs.wrapping_sub(rhs))
+            }
+            Self::And => {
+                let lhs: i32 = lhs.try_into().expect("and operand overflow!");
+                let rhs: i32 = rhs.try_into().expect("and operand overflow!");
+                i64::from(lhs & rhs)
```
koute, line 3492:
```suggestion
                lhs & rhs
```

I like that you're being strict here! But the casts here for the bitwise ops are not really necessary due to the very nature of how bitwise ops work.

If you don't want to think about it too much (for some operations these kinds of properties are not immediately obvious), here's a [Kani](https://model-checking.github.io/kani/) proof:

```rust
#[kani::proof]
fn foobar() {
    let a: i32 = kani::any();
    let b: i32 = kani::any();

    let result32 = a & b;
    let result64 = i64::from(a) & i64::from(b);
    assert_eq!(i64::from(result32), result64);
    assert_eq!(result32, result64 as i32);
}
```

This passes:

```
RESULTS:
Check 1: foobar.assertion.1
	 - Status: SUCCESS
	 - Description: "assertion failed: i64::from(result32) == result64"
	 - Location: src/main.rs:15:5 in function foobar

Check 2: foobar.assertion.2
	 - Status: SUCCESS
	 - Description: "assertion failed: result32 == result64 as i32"
	 - Location: src/main.rs:16:5 in function foobar
```

If you're not sure whether a piece of code you're writing is correct or not then Kani is very useful tool to be able to prove it.

Of course you can never have too many extra asserts, and in this case (at least for now) `OperationKind::And` should only ever be materialized when we're compiling 32-bit code, so having a check that `lhs` and `rhs` fits in 32-bit is good. So I think a better way to do this would be to have `assert!(i32::try_from(lhs).is_some())` here to make sure the inputs are 32-bit, and just use `lhs & rhs` to calculate the result without any casts.

Why I'm writing this: currently we have `OperationKind::And` (which will be materialized when we're compiling 32-bit) and `OperationKind::And64` (which will be materialized when we're compiling 64-bit). But this is *probably* unnecessary, and we could simplify the code and only have `OperationKind::And` which works correctly on both.

----

So, as a side note, if we look at all of the `OperationKind` we can split them into three categories:

- Those which will only be materialized when we're compiling a 32-bit program (e.g. `OperationKind::And`)
- Those which will only be materialized when we're compiling a 64-bit program (e.g. `OperationKind::And64`)
- Those which will be materialized when we're compiling both 32-bit and 64-bit programs (e.g. `OperationKind::Add`)

So if you look at the [RISC-V spec](https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf) and how it defines the 32-bit and 64-bit targets here's how it does it:

- On 32-bit instructions naturally operate on 32-bits.
- On 64-bit instructions naturally operate on 64-bits (while having the same opcode as on 32-bit, which is why Cyril made all of those extra 64-bit `OperationKind`s)
- On 64-bit there are a few extra instructions which are essentially the 32-bit version of the instruction, just encoded with a different opcode.

So, for example, let's take the ADD RISC-V instruction:

- On 32-bit when we see ADD we'll materialize `OperationKind::Add` (we do this substitution immediately when we parse RISC-V code in `riscv.rs`)
- On 64-bit when we see ADD we'll materialize `OperationKind::Add64`
- On 64-bit when we see ADDW (this is the extra 64-bit only instruction which emulates the original 32-bit ADD) we'll materialize `OperationKind::Add`

So in this case what's a correct implementation of `OperationKind::Add`? Here's what the RISC-V spec says should happen when we're on a 64-bit target:

> ADDW and SUBW are RV64I-only instructions that are defined analogously to ADD and SUB
but operate on 32-bit values and produce signed 32-bit results. Overflows are ignored, and the low 32-bits of the result is sign-extended to 64-bits and written to the destination register.

So in case of `OperationKind::Add` the correct implementation would be something like this:

```rust
lhs.wrapping_add(rhs) as i32 as i64
```

So we do an addition in 64-bits. We do a wrapping addition ("overflows are ignored") then we cast to i32 to grab the lower 32-bits ("the low 32-bits of the result") and we cast them up to 64-bit again ("is sign-extended to 64-bits").

And here's a Kani proof that it's also correct for 32-bit programs:

```rust
#[kani::proof]
fn foobar() {
    let a32: i32 = kani::any();
    let a64 = i64::from(a32);
    let b32: i32 = kani::any();
    let b64 = i64::from(b32);

    let result32 = a32.wrapping_add(b32);
    let result64 = a64.wrapping_add(b64) as i32 as i64;
    assert_eq!(i64::from(result32), result64);
    assert_eq!(result32, result64 as i32);
}
```

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3621,14 +3699,14 @@ impl BlockRegs {
             regs: Reg::ALL.map(|reg| RegValue::OutputReg {
                 reg,
                 source_block,
-                bits_used: u32::MAX,
+                bits_used: u64::from(u32::MAX),
```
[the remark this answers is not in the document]
aman4150: Agreed, but the change is not expected to handle 64-bit binary at the moment. This should get fixed in the follow change

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -1794,7 +1802,10 @@ fn convert_instruction(
                 // The optimizer can take care of this later, but doing it early here is more efficient.
                 emit(InstExt::Basic(BasicInst::LoadImmediate {
                     dst,
-                    imm: OperationKind::from(kind).apply_const(0, imm),
+                    imm: OperationKind::from(kind)
+                        .apply_const(0, i64::from(imm))
```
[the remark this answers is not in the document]
aman4150: Yes that PR takes care of integer upcast based on if the operand is i32 or u32, I will add a helper function.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: I agree cast is not necessary for bitwise ops :) but my intention was to be consistent with operand checks.

Thanks for introducing me to the Kani proof!

aman4150: unless you've strong opinions about adding bound checks and casts for bitwise ops, I would prefer to keep it.

Please let me know :)

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3590,15 +3665,18 @@ impl RegValue {
                 dst,
                 target: AnyTarget::Data(target),
             }),
-            RegValue::Constant(imm) => Some(BasicInst::LoadImmediate { dst, imm }),
+            RegValue::Constant(imm) => {
+                let imm = i32::try_from(imm).expect("Load immediate operand overflow! {}");
```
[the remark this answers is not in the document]
aman4150: thanks, will remove it in a follow up revision.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -6399,7 +6477,11 @@ fn emit_code(
                             if is_optimized {
                                 unreachable!("internal error: instruction with only constant operands: {op:?}")
                             } else {
-                                I::load_imm(dst, OperationKind::from(kind).apply_const(src1 as i32, src2 as i32) as u32)
+                                let imm: u32 = OperationKind::from(kind)
+                                    .apply_const(i64::from(src1 as i32), i64::from(src2 as i32))
+                                    .try_into()
+                                    .expect("Load immediate overflow!");
```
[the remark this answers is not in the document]
aman4150: Sure, will fix it.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: Discussed this offline, we can get rid of redundant OperationKind (32 bit version of Equality and bitwise) later, but for now we can keep the implementation as-is.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3590,15 +3665,18 @@ impl RegValue {
...
             _ => None,
         }
     }
 
-    fn bits_used(self) -> u32 {
+    fn bits_used(self) -> u64 {
         match self {
             RegValue::InputReg(..) | RegValue::CodeAddress(..) | RegValue::DataAddress(..) => !0,
-            RegValue::Constant(value) => value as u32,
+            RegValue::Constant(value) => value as u64,
```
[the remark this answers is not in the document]
aman4150: Makes sense, will fix it.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3958,15 +4036,15 @@ impl BlockRegs {
...
                 kind: LoadKind::U16, dst, ..
             }
             | BasicInst::LoadIndirect {
                 kind: LoadKind::U16, dst, ..
             } => {
-                self.set_reg_unknown(dst, unknown_counter, u32::from(u16::MAX));
+                self.set_reg_unknown(dst, unknown_counter, u64::from(u16::MAX));
             }
             _ => {
                 for dst in instruction.dst_mask(imports) {
```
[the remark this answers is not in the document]
aman4150: if `bits_used` member function of `RegValue` enum does the right thing, then it doesn't matter if the setter respects register width or not. Do you agree?

koute (approved):

PR #179, opened 2024-10-03
fuzz: interpreter

subotic:


fuzz/fuzz_targets/fuzz_interpreter.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use libfuzzer_sys::fuzz_target;
+use polkavm::Engine;
+use polkavm::InterruptKind;
+use polkavm::ModuleConfig;
+use polkavm::ProgramBlob;
+use polkavm::ProgramCounter;
+
+fn harness(_data: &[u8]) {
+    // configure the polkavm engine
```
jarkkojs, line 11: Why this comment is useful, or how anyone could be misled otherwise?

fuzz/fuzz_targets/fuzz_interpreter.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    // create a polkavm module from the engine, module config, and program blob
+    let module = polkavm::Module::from_blob(&engine, &module_config, blob).unwrap();
+
+    let initial_pc = ProgramCounter(0);
+    let mut final_pc = initial_pc;
+
+    // instantiate the module and run it
+    let mut instance = module.instantiate().unwrap();
+    instance.set_gas(1000000);
```
jarkkojs, line 39: A magic number. Should be const defined before the function in the files namespace.

koute: Actually, here it is fine. (: Let's not cargo cult; just because this is a number it doesn't mean it's "magic". :P

In general my policy for numbers is that you put it in a dedicated const if:
a) you need to use the same number multiple times *and* if it's changed then it will have to be changed everywhere (otherwise the code will be incorrect)
b) giving the number a name is useful/gives extra information to the reader.

In this particular case it's neither. It's not (a) because it's only used once, and it's not (b) because if you do:

```rust
const GAS: u64 = 1000000;
instance.set_gas(GAS);
```

it's *completely* useless and it doesn't give you any new information nor make the code any more readable (if anything it makes it worse as it just bloats the line count without any benefit). You already know this is gas because, well, it's used in the `set_gas` function, so the extra const is just noise.

jarkkojs: OK, yeah I guess you have a point!

jarkkojs: So how do I come up with an appropriate number to supply to `set_gas()`? [meant only as out-of-band question]

jarkkojs: [I neither know what cargo cult is for what that is worth :-) ]

Issue #178, opened 2024-09-27
Support upstream toolchain


koute:
We're currently using [our own fork of rustc](https://github.com/paritytech/rustc-rv32e-toolchain/) to build guest binaries, but in theory it should be possible to now use a purely upstream build of rustc since rv32e support got merged into LLVM.

Copy-paste this to `guest-programs/riscv32emac-unknown-none-polkavm.json`:

```
{
  "arch": "riscv32",
  "cpu": "generic-rv32",
  "crt-objects-fallback": "false",
  "data-layout": "e-m:e-p:32:32-i64:64-n32-S32",
  "eh-frame-header": false,
  "emit-debug-gdb-scripts": false,
  "features": "+e,+m,+a,+c,+lui-addi-fusion,+fast-unaligned-access,+xtheadcondmov",
  "linker": "rust-lld",
  "linker-flavor": "ld.lld",
  "llvm-abiname": "ilp32e",
  "llvm-target": "riscv32",
  "max-atomic-width": 32,
  "panic-strategy": "abort",
  "relocation-model": "pie",
  "target-pointer-width": "32",
  "singlethread": true,
  "pre-link-args": {
    "ld": [
      "--emit-relocs",
      "--unique"
    ]
  },
  "env": "polkavm"
}
```

Then go into `guest-programs/build-benchmarks.sh` and apply this:

```
diff --git a/guest-programs/build-benchmarks.sh b/guest-programs/build-benchmarks.sh
index e5435c5..966b28b 100755
--- a/guest-programs/build-benchmarks.sh
+++ b/guest-programs/build-benchmarks.sh
@@ -77,9 +77,9 @@ function build_benchmark() {
 
     if [ "${RV32E_TOOLCHAIN:-}" != "" ]; then
         echo "> Building: '$1' (polkavm)"
-        RUSTFLAGS="-C target-feature=+lui-addi-fusion,+fast-unaligned-access,+xtheadcondmov,+c -C relocation-model=pie -C link-arg=--emit-relocs -C link-arg=--unique $extra_flags" rustup run $RV32E_TOOLCHAIN cargo build -q --release --bin $1 -p $1
+        RUSTFLAGS="$extra_flags" rustup run nightly-2024-07-10 cargo build -Z build-std=core,alloc --target "$current_dir/riscv32emac-unknown-none-polkavm.json" -q --release --bin $1 -p $1
         cd ..
[...]

PR #184, opened 2024-10-08
refactor: change to the official riscv32emac toolchain

jarkkojs:
Change from the Parity's rv32e fork to the official rustc toolchain.

Makes sure that also PolkaVM exports get relocation entry with the '--relocatable' parameters. Otherwise the linker will only add the relocs, whic are referenced by the code directly.

Flag DWARF processing under the feature flag 'dwarf' for the moment.

Link: https://github.com/paritytech/rustc-rv32e-toolchain/
Closes: #178

### DIFF
--- a/crates/polkavm-linker/Cargo.toml
+++ b/crates/polkavm-linker/Cargo.toml
@@ -1,3 +1,7 @@
+[features]
+default = []
+dwarf = []
+
 [package]
 name = "polkavm-linker"
 version.workspace = true
--- a/crates/polkavm-linker/src/dwarf.rs
+++ b/crates/polkavm-linker/src/dwarf.rs
@@ -1815,12 +1815,10 @@ struct LineEntry {
     location: SourceCodeLocation,
 }
 
+/// Represents a DWARF section. This is not the most efficient representation,
+/// but it's invariant to any transformations that might be applied to the code.
 #[derive(Default)]
-pub(crate) struct DwarfInfo {
-    // This is not the most efficient representation, but it's invariant
-    // to any transformations that might be applied to the code.
-    pub location_map: HashMap<SectionTarget, Arc<[Location]>>,
-}
+pub(crate) struct DwarfInfo(pub HashMap<SectionTarget, Arc<[Location]>>);
 
 struct AttributeValue<R>
 where
@@ -2241,7 +2239,7 @@ where
     Ok(None)
 }
 
-pub(crate) fn load_dwarf<H>(
+pub(crate) fn load<H>(
     string_cache: &mut StringCache,
     elf: &Elf<H>,
     relocations: &BTreeMap<SectionTarget, RelocationKind>,
@@ -2329,5 +2327,5 @@ where
     };
 
     let location_map = walker.run()?;
-    Ok(DwarfInfo { location_map })
+    Ok(DwarfInfo(location_map))
 }
--- a/crates/polkavm-linker/src/program_from_elf.rs
+++ b/crates/polkavm-linker/src/program_from_elf.rs
@@ -8311,8 +8311,9 @@ where
     let mut location_map: HashMap<SectionTarget, Arc<[Location]>> = HashMap::new();
     if !config.strip {
         let mut string_cache = crate::utils::StringCache::default();
-        let dwarf_info = crate::dwarf::load_dwarf(&mut string_cache, &elf, &relocations)?;
-        location_map = dwarf_info.location_map;
+        if cfg!(feature = "dwarf") {
+            location_map = (crate::dwarf::load(&mut string_cache, &elf, &relocations)?).0;
+        }
 
         // If there is no DWARF info present try to use the symbol table as a fallback.
         for (source, name) in parse_function_symbols(&elf)? {
--- a/guest-programs/.cargo/config.toml
+++ b/guest-programs/.cargo/config.toml
@@ -1,7 +1,7 @@
 [build]
-target = "riscv32ema-unknown-none-elf"
+target = "riscv32emac-unknown-none-polkavm"
 
-[target.riscv32ema-unknown-none-elf]
+[target.riscv32emac-unknown-none-polkavm]
 rustflags = [
     "-C", "relocation-model=pie",
     "-C", "link-arg=--emit-relocs",
--- a/guest-programs/build-benchmarks.sh
+++ b/guest-programs/build-benchmarks.sh
@@ -1,13 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
 cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
 
 source ../ci/jobs/detect-or-install-riscv-toolchain.sh
-
-if [ "${RV32E_TOOLCHAIN:-}" == "" ]; then
-    echo "WARN: rv32e toolchain is missing; PolkaVM binaries won't be built!"
-fi
 
 TOOLCHAIN_VERSION="1.75.0"
 
@@ -71,17 +67,27 @@ if [ "${SOLANA_PLATFORM_TOOLS_DIR:-}" == "" ]; then
     esac
 fi
 
+build_polkavm() {
+    echo "> Building: '$1' (polkavm)"
+
+    RUSTFLAGS="$extra_flags" cargo build -Z build-std=core,alloc \
+        --target "$PWD/riscv32emac-unknown-none-polkavm.json" \
+        -q --release --bin $1 -p $1
+
+    pushd ..
+
+    cargo run -q -p polkatool link \
+        --run-only-if-newer guest-programs/target/riscv32emac-unknown-none-polkavm/release/$1 \
+        -o guest-programs/target/riscv32emac-unknown-none-polkavm/release/$1.polkavm
+
+    popd
+}
+
 function build_benchmark() {
-    current_dir=$(pwd)
     extra_flags="${extra_flags:-}"
 
-    if [ "${RV32E_TOOLCHAIN:-}" != "" ]; then
-        echo "> Building: '$1' (polkavm)"
-        RUSTFLAGS="-C target-feature=+lui-addi-fusion,+fast-unaligned-access,+xtheadcondmov,+c -C relocation-model=pie -C link-arg=--emit-relocs -C link-arg=--unique $extra_flags" rustup run $RV32E_TOOLCHAIN cargo build -q --release --bin $1 -p $1
-        cd ..
-        cargo run -q -p polkatool link --run-only-if-newer guest-programs/target/riscv32ema-unknown-none-elf/release/$1 -o guest-programs/target/riscv32ema-unknown-none-elf/release/$1.polkavm
-        cd $current_dir
-    fi
+    # Unconditional build:
+    build_polkavm $1
 
     if [ "${BUILD_WASM}" == "1" ]; then
         echo "> Building: '$1' (wasm)"
--- a/guest-programs/riscv32emac-unknown-none-polkavm.json
+++ b/guest-programs/riscv32emac-unknown-none-polkavm.json
@@ -1,0 +1,26 @@
+{
+  "arch": "riscv32",
+  "cpu": "generic-rv32",
+  "crt-objects-fallback": "false",
+  "data-layout": "e-m:e-p:32:32-i64:64-n32-S32",
+  "eh-frame-header": false,
+  "emit-debug-gdb-scripts": false,
+  "features": "+e,+m,+a,+c,+lui-addi-fusion,+fast-unaligned-access,+xtheadcondmov",
+  "linker": "rust-lld",
+  "linker-flavor": "ld.lld",
+  "llvm-abiname": "ilp32e",
+  "llvm-target": "riscv32",
+  "max-atomic-width": 32,
+  "panic-strategy": "abort",
+  "relocation-model": "pie",
+  "target-pointer-width": "32",
+  "singlethread": true,
+  "pre-link-args": {
+    "ld": [
+      "--emit-relocs",
+      "--unique",
+      "--relocatable"
+    ]
+  },
+  "env": "polkavm"
+}
--- a/guest-programs/rust-toolchain.toml
+++ b/guest-programs/rust-toolchain.toml
@@ -1,2 +1,3 @@
 [toolchain]
-channel = "riscv32em-nightly-2024-01-05-r0-x86_64-unknown-linux-gnu"
+channel = "nightly-2024-07-10"
+components = ["rust-src"]


crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -8311,8 +8311,9 @@ where
     let mut location_map: HashMap<SectionTarget, Arc<[Location]>> = HashMap::new();
     if !config.strip {
         let mut string_cache = crate::utils::StringCache::default();
-        let dwarf_info = crate::dwarf::load_dwarf(&mut string_cache, &elf, &relocations)?;
-        location_map = dwarf_info.location_map;
+        if cfg!(feature = "dwarf") {
+            location_map = (crate::dwarf::load(&mut string_cache, &elf, &relocations)?).0;
+        }
```
athei, line 8316: Why would it be a feature flag? Let's just make it a runtime option.

guest-programs/build-benchmarks.sh
```diff
@@ -1,14 +1,10 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
 cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
 
 source ../ci/jobs/detect-or-install-riscv-toolchain.sh
```
koute, line 6: This line should be deleted, and that script should be deleted altogether.

crates/polkavm-linker/Cargo.toml
```diff
@@ -1,3 +1,7 @@
+[features]
+default = []
+dwarf = []
```
koute, line 3: We shouldn't have any features here (DWARF support should be always compiled in), and we already have a runtime flag which makes the linker ignore DWARF (the `--strip` flag in polkatool/`set_strip` in linker's config struct)

crates/polkavm-linker/Cargo.toml
```diff
@@ -1,3 +1,7 @@
+[features]
+default = []
```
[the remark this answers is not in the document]
jarkkojs: By evolution and re-learning Rust, i.e. I tried to recall how feature flags where defined and this is only cruft. I don't mind removing it.

crates/polkavm-linker/src/program_from_elf.rs
jarkkojs: Sorry, define "runtime option".

guest-programs/build-benchmarks.sh
jarkkojs: `#!/usr/bin/env bash` is POSIX cross-compatible, and not just in theory, there is variance with `/bin/bash` and `/usr/bin/bash`. Technically (if script was kept) this should be a separate fix.

But since your request was to remove this, and the issue defined running this script as the test case. How do I test it afterwards?

crates/polkavm-linker/Cargo.toml
jarkkojs: So in today's meeting we made the plan that since DWARF parsing is a separate issue from relocation, and it would be better to get the official toolchain in use, it is flagged and a new issue is created to fix it.

guest-programs/.cargo/config.toml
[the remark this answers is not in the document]
jarkkojs: DOne

guest-programs/build-benchmarks.sh
jarkkojs: So I keep this open until the pull request is final. I updated the PR because it already fixes two issues in DWARF code.

Fixed: overlap bug
Open: how to deal DW_AT_specification with zero offset

koute: > `#!/usr/bin/env bash` is POSIX cross-compatible, and not just in theory, there is variance with `/bin/bash` and `/usr/bin/bash`. Technically (if script was kept) this should be a separate fix.

This wasn't the line I was talking about. (: In general, I don't care which one it is, as long as it works. I always tend to use `/bin/bash` by default because that seems to work everywhere for me, but if you use some weird distro where it doesn't then feel free to go wild and change it.

I highlighed this line (the line immediately above my comment):

```
source ../ci/jobs/detect-or-install-riscv-toolchain.sh
```

If we're switching to the upstream toolchain then we don't need this line, so it should be deleted along with the script.

jarkkojs:
Not worth commenting just yet. There's still a few issues but all are related just twists in DWARF data, which did not exist when compiled with Parity's toolchain.

Summary:
- Most of the fixes already in place might need some tweaking but mostly are correctly identified bugs.
- There's one issue at least left that I simply did not yet investigate properly but can be addressed probably just by looking the binary with rz-bin.
- @athei asked to remove .cargo/config.toml but I'm not sure about this but open for feedback. @koute on the other asked to remove build-bechmark.sh. So some *overall* conclusions and consensus "do this with build shenanigans" would be nice here. No focus on this before I've fixed the remaining DWARF issues.

So I do a temporary work branch to finalize this pull request, thus does not yet much sense to give any detailed feedback on the code itself. I've tried to add some rz-bin output to each commit so that it describes clearly the situation in the binary and kind of helps to understand from those grounds what the code change does (as otherwise these make no sense at all).

jarkkojs:
So this iteration #2 of the pull request.

Open:

```
.github/workflows/rust.yml:      run: ./ci/jobs/detect-or-install-riscv-toolchain.sh
.github/workflows/rust.yml:      run: ./ci/jobs/detect-or-install-riscv-toolchain.sh
ci/jobs/build-guests.sh:source ./ci/jobs/detect-or-install-riscv-toolchain.sh
ci/jobs/clippy.sh:source ./ci/jobs/detect-or-install-riscv-toolchain.sh
ci/jobs/rustfmt.sh:source ./ci/jobs/detect-or-install-riscv-toolchain.sh
ci/run-all-tests.sh:source ./ci/jobs/detect-or-install-riscv-toolchain.sh
guest-programs/build-examples.sh:source ../ci/jobs/detect-or-install-riscv-toolchain.sh
```

I think it is better to have a checkpoint (feedback) first before fixing up these (in a separate commit).

Changes to `dwarfs.rs` are bottom because I wanted to test them also with the previous toolchain. With these changes now also `build-benchmark.sh` passes with the new toolchain.

crates/polkavm-linker/src/dwarf.rs
```diff
@@ -1455,9 +1466,7 @@ where
                 };
 
                 if decl_location.is_some() && name.is_none() {
-                    return Err(ProgramFromElfError::other(
-                        "failed to process DWARF: subprogram has a decl location but no name",
-                    ));
+                    log::warn!("failed to process DWARF: subprogram has a decl location but no name");
```
koute, line 1469: Hm, so do you know if these are due to broken DWARF, or if this is just "normal"? If it's normal then this log should probably be downgraded to `debug` or `trace`.

But if you don't know then it's fine as-is.

(This applies to every `warn!` you've added here.)

koute (commented):
Looks mostly good to me!

We can merge this once we fix clippy.

crates/polkavm-linker/src/dwarf.rs
jarkkojs: We need to figure these out (as discussed) in use. Right now I think this right action to take. Features and bugs really make them resolve themselves.

In the long-run I'd also consider a split to "subprogram" and "subroutine" phases as subprograms can be parsed without subroutines and with experimental change this really seems to make the whole so much more clean.

Subprograms alone have their own special cases such as out-of-line inlines (which are not actually inlines) but they never have subroutine parents. Only subroutines point to the subprograms. I think this would help to sort out many of these. I did not want to include this change just yet because it goes way beyond toolchain change.

It's like the next thing when something major specific to DWARF needs to be done. So not dealing with all of this right now (the ones that made sense to deal are separate commits) but I have ideas and plans how to deal with them when we need to :-) 

So with this I think we can resolve this (in the context of this pull request).

jarkkojs:
So I there was issue I was looking into on removing that CI script see first this temp diff:

```
diff --git a/crates/polkavm-linker/src/program_from_elf.rs b/crates/polkavm-linker/src/program_from_elf.rs
index 0c60355..d18dcff 100644
--- a/crates/polkavm-linker/src/program_from_elf.rs
+++ b/crates/polkavm-linker/src/program_from_elf.rs
@@ -1408,6 +1408,7 @@ where
     let _ = b.read(target.offset as usize)?;
 
     let version = b.read_byte()?;
+    log::debug!("section {} {}", section.name(), version);
     if version != 1 && version != 2 {
         return Err(format!("unsupported extern metadata version: '{version}' (expected '1' or '2')"));
     }
diff --git a/guest-programs/build-examples.sh b/guest-programs/build-examples.sh
index 39542a3..0e00312 100755
--- a/guest-programs/build-examples.sh
+++ b/guest-programs/build-examples.sh
@@ -12,13 +12,22 @@ fi
 
 function build_example () {
     output_path="output/$1.polkavm"
-    current_dir=$(pwd)
 
     echo "> Building: '$1' (-> $output_path)"
-    RUSTFLAGS="-C target-feature=+c -C relocation-model=pie -C link-arg=--emit-relocs -C link-arg=--unique --remap-path-prefix=$(pwd)= --remap-path-prefix=$HOME=~" cargo build -q --release --bin $1 -p $1
-    cd ..
-    cargo run -q -p polkatool link --run-only-if-newer -s guest-programs/target/riscv32ema-unknown-none-elf/release/$1 -o guest-programs/$output_path
-    cd $current_dir
+
+    RUSTFLAGS="--remap-path-prefix=$(pwd)= --remap-path-prefix=$HOME=~" \
+    cargo build  \
+        -Z build-std=core,alloc \
+        --target "$PWD/riscv32emac-unknown-none-polkavm.json" \
+        -q --release --bin $1 -p $1
+
+    pushd ..
+
+    cargo run -q -p polkatool link \
+        --run-only-if-newer guest-programs/target/riscv32emac-unknown-none-polkavm/release/$1 \
+        -o guest-programs/$output_path
+
+    popd
 }
 
 build_example "example-hello-world"
```

The result:

```
❯ RUST_BACKTRACE=1 RUST_LOG="polkavm_linker::program_from_elf=debug" guest-programs/build-examples.sh   
> Building: 'example-hello-world' (-> output/example-hello-world.polkavm)
~/work/github.com/jarkko/polkavm ~/work/github.com/jarkko/polkavm/guest-programs
[DEBUG polkavm_linker::program_from_elf] section .polkavm_metadata 1
[DEBUG polkavm_linker::program_from_elf] section .polkavm_metadata 0
ERROR: failed to link "guest-programs/target/riscv32emac-unknown-none-polkavm/release/example-hello-world": failed to parse extern metadata: unsupported extern metadata version: '0' (expected '1' or '2')
```

I sanity checked again with old toolchain that my DWARF changes could not possibly cause any of this and they do not. `build-examples.sh` works perfectly with all of them. I also checked what happens with benchmarks and with them the version is 1 for both exports.

I looked at also `polkavm-derive` crate and found the generators and it version 1 is defined for `METADATA`.

I keep also Friday off work (like I did Wed) to keep my hours in balance so looking into this next Monday.

PS. A more standard convention would to use ELF note section (e.g. `.note.polkavm`) for this. Just saying this in completeness no strong push ATM :-)

jarkkojs:
I'll check on Monday what the binary dumps comparing exports section (or whatever it was called) with old/new toolchain (dunno why did not came to mind Thu).

guest-programs/build-examples.sh
```diff
@@ -5,20 +5,26 @@ cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
...
-    current_dir=$(pwd)
 
     echo "> Building: '$1' (-> $output_path)"
-    RUSTFLAGS="-C target-feature=+c -C relocation-model=pie -C link-arg=--emit-relocs -C link-arg=--unique --remap-path-prefix=$(pwd)= --remap-path-prefix=$HOME=~" cargo build -q --release --bin $1 -p $1
-    cd ..
-    cargo run -q -p polkatool link --run-only-if-newer -s guest-programs/target/riscv32ema-unknown-none-elf/release/$1 -o guest-programs/$output_path
-    cd $current_dir
+
+    local target="riscv32emac-unknown-none-polkavm"
+    local extra_flags="-C target-feature=+c -C relocation-model=pie -C link-arg=--emit-relocs -C link-arg=--unique --remap-path-prefix=$(pwd)= --remap-path-prefix=$HOME=~"
```
athei, line 14: Do we still need all of those args? They seem to be already set by the target.

guest-programs/build-examples.sh
```diff
@@ -5,20 +5,26 @@ cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
...
+
+    RUSTFLAGS="$extra_flags" cargo build  \
+        -Z build-std=core,alloc \
+        --target "$PWD/$target.json" \
+        -q --release --bin $1 -p $1
+
+    pushd ..
+
+    cargo run -q -p polkatool link \
+        --run-only-if-newer -s guest-programs/target/$target/release/$1 \
```
athei, line 24: `-s` means we are stripping and hence skipping DWARF in the linker. Is that intended?

guest-programs/build-examples.sh
jarkkojs: Probably not, but did not yet remove it given inconsistent data generated to the exports section. Consider this commit WiP up until the issue root caused. I'll keep this open tho.

guest-programs/build-examples.sh
jarkkojs: Probably/possibly not. I'll check this. I had no idea what the option is doing and it existed before the change.

To be more formal, let's compare it to `build-benchmark.sh`:

```
    cargo run -q -p polkatool link \
        --run-only-if-newer guest-programs/target/riscv32emac-unknown-none-polkavm/release/$1 \
        -o guest-programs/target/riscv32emac-unknown-none-polkavm/release/$1.polkavm
```

Yep, it's likely that it is not needed.

guest-programs/build-examples.sh
jarkkojs: Let's enumerate and evaluate:

1. `-C target-feature=+c`
2. `-C relocation-model=pie`
3. `-C link-arg=--emit-relocs`
4. `-C link-arg=--unique`
5. `--remap-path-prefix=$(pwd)= --remap-path-prefix=$HOME=`

2, 3 and 4 are dead obvious removal but I'm lost with 1 and 5. I don't even know what they are, and what they were meant to do originally. Perhaps @koute can bring context here? Would be educational to know even they would get scraped :-)

I'm also quite confused, informal googling gives me the idea that 5 should be in fact `--remap-path-prefix-from=$(pwd)= --remap-path-prefix-to=$HOME=`. Still, causes no compilation errors.

jarkkojs:
zZz and wait for input... I wrote on Sunday (and sort of used half of my Monday, also rebased the patch set and fixed merge conflicts and hit my head wall with this issue under discussion) extra patch that simply allows to optionally pick benchmark with a command-line parameter. It's easier to fix build and linking issues because you don't have to comment out stuff by hand from `build-benchmarks.sh`.

guest-programs/build-examples.sh
koute: (1) is enabling the C extension because the original toolchain didn't have it enabled by default; on the upstream toolchain that should be unnecessary (because the target JSON file should have it enabled by default).

(5) replaces paths in the debug info; by default you'd get paths like `/home/user/.cargo/...` hardcoding the username of whoever compiled the file and `/path/to/checkout/polkavm` hardcoding the exact place where you've checked out the repo, and we don't want that (ideally we want builds to be bit-by-bit reproducible), so this snips these paths and makes them relative

guest-programs/build-examples.sh
koute: This is intended here in this script because the resulting blob is checked into the repo.

jarkkojs:
After fixing up the parameters in `build-examples.sh` according to the comments:

```bash
❯ guest-programs/build-examples.sh 
> Building: 'example-hello-world' (-> output/example-hello-world.polkavm)
~/work/github.com/jarkko/polkavm ~/work/github.com/jarkko/polkavm/guest-programs
ERROR: failed to link "guest-programs/target/riscv32emac-unknown-none-polkavm/release/example-hello-world": failed to parse extern metadata: unsupported extern metadata version: '0' (expected '1' or '2')
```

So the problem still persists.

jarkkojs:
@koute here's the dumps.

New toolchain:

```bash
 objdump -s -j .polkavm_exports ./guest-programs/target/riscv32emac-unknown-none-polkavm/release/example-hello-world

./guest-programs/target/riscv32emac-unknown-none-polkavm/release/example-hello-world:     file format elf32-little

Contents of section .polkavm_exports:
 0000 01000000 00000000 00                 .........       
```

Old toolchain:

```bash
❯ objdump -s -j .polkavm_exports ./guest-programs/target/riscv32ema-unknown-none-elf/release/example-hello-world 

./guest-programs/target/riscv32ema-unknown-none-elf/release/example-hello-world:     file format elf32-little

Contents of section .polkavm_exports:
 0000 01942101 00641101 00                 ..!..d...       
```

jarkkojs:
Just documenting RISC-V relocations.

Both truths hold also for RISC-V when it comes to relocations:

1. `R_RISCV_32`: **S +A** (also known as *absolute relocation*)
2. `R_RISCV_PC32`: **S+A-P** (also known as *relative relocation*)

In the latter **P** is location of the instruction/whatever in the memory which is computed by the formula `sh_addr + r_offset`. This is 1:1 exact match how 386 relocations work in ELF.

Relocations are not universally independent on section addresses. Only the absolute relocations are like this. This good to keep in mind and also that it depends case by case.

jarkkojs:
Follow-up: https://github.com/paritytech/polkavm/issues/199

jarkkojs:
Looking this next time on Monday, if it is still open but I'd hope that since I created follow-up issues and this passes the CI, it'd be a great time to merge it, in order to get a milestone (and move forward with #199).

jarkkojs:
Please use "rebase and merge" instead of "squash and merge" so that my splits don't go wasted, and will in future useful for `git bisect` and `git blame -s` when seeking for failing commits. The default was "squash and merge", thus the remark.

koute:
> Please use "rebase and merge" instead of "squash and merge" so that my splits don't go wasted

If the commits are nice I always use rebase and merge (and in general I personally prefer a rebase-based workflow myself), but FYI this is not true for Parity at large (if you're going to contribute to `polkadot-sdk` then there the policy is to never rebase, and to always squash).

ci/jobs/build-guests.sh
```diff
@@ -1,13 +1,11 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
-cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
+
+cd "${0%/*}/"
```
koute, line 5: Is there actually a difference between what I've been doing previously and this approach? Is this one superior?

guest-programs/build-benchmarks.sh
```diff
@@ -71,17 +65,28 @@ if [ "${SOLANA_PLATFORM_TOOLS_DIR:-}" == "" ]; then
...
 
+build_polkavm() {
+    echo "> Building: '$1' (polkavm)"
+
+    RUSTFLAGS="$extra_flags" cargo build  \
+        -Z build-std=core,alloc \
+        --target "$PWD/riscv32emac-unknown-none-polkavm.json" \
+        -q --release --bin $1 -p $1
+
+    pushd ..
```
koute, line 76: Can you silence the `pushd`/`popd`? It spams the console with the paths, which is just unnecessary noise.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -1410,9 +1414,18 @@ where
...
     let _ = b.read(target.offset as usize)?;
 
-    let version = b.read_byte()?;
+    let mut version = b.read_byte()?;
+
+    // FIXME: `.polkavm_metadata` section is padded with 0 when it starts with
+    // an import for unknown reason.
+    if version == 0 {
+        version = b.read_byte()?;
+    }
```
koute, line 1427: This should be deleted.

The issue seems to be due to me not specifying a proper repr on the metadata structs; here's a patch with a fix:

```diff
diff --git a/crates/polkavm-derive-impl/src/abi_support_impl/common_private.rs b/crates/polkavm-derive-impl/src/abi_support_impl/common_private.rs
index 6c06da6..9dbd7aa 100644
--- a/crates/polkavm-derive-impl/src/abi_support_impl/common_private.rs
+++ b/crates/polkavm-derive-impl/src/abi_support_impl/common_private.rs
@@ -395,7 +395,7 @@ impl PackReturnTy for (Reg, Reg) {
 pub struct MetadataPointer(pub *const u8);
 unsafe impl Sync for MetadataPointer {}
 
-#[repr(packed)]
+#[repr(C, packed(1))]
 pub struct ExternMetadataV1 {
     pub version: u8,
     pub flags: u32,
@@ -405,7 +405,7 @@ pub struct ExternMetadataV1 {
     pub output_regs: u8,
 }
 
-#[repr(packed)]
+#[repr(C, packed(1))]
 pub struct ExternMetadataV2 {
     pub version: u8,
     pub flags: u32,
```

guest-programs/bench-prime-sieve/src/unrolled.rs
```diff
@@ -91,6 +91,7 @@ fn reinterpret_slice_mut_u64_u8(words: &mut [u64]) -> &mut [u8] {
 /// Storage structure implementing standard linear bit storage, but with a hybrid bit setting strategy:
 /// - dense resetting for small skip factors
 /// - sparse resetting for larger skip factors
+///
```
koute, line 94: Is this due to `cargo fmt`?

jarkkojs: Not that one. These were reported by the clippy script.

ci/jobs/build-guests.sh
jarkkojs: It is heck a lot more cleaner and compact at least.

guest-programs/build-benchmarks.sh
jarkkojs: +1

I'll look into this on Monday.

jarkkojs:
> > Please use "rebase and merge" instead of "squash and merge" so that my splits don't go wasted
> 
> If the commits are nice I always use rebase and merge (and in general I personally prefer a rebase-based workflow myself), but FYI this is not true for Parity at large (if you're going to contribute to `polkadot-sdk` then there the policy is to never rebase, and to always squash).

OK cool it was only what Github seems to recommend me as default (again apparently) :-)

crates/polkavm-linker/src/program_from_elf.rs
jarkkojs: I agree but since it took me so long time to find out the bug, and find some way to address it, I'd see it would remove from us valuable information to remove my workaround. Especially since the workaround has been marked with `FIXME`, i.e. everyone can see it was not final, and since I already created #199.

So rather than overwriting workaround (while not disagreeing with the actual fix) I'm open here for two alternatives:

1. I'll add a fix on top of this closing #199.
2. We'll finalize this PR and then close #199.

I.e. I see my workaround and your final fix both valuable as "documentation", and acting otherwise would erase useful historical data.

guest-programs/bench-prime-sieve/src/unrolled.rs
jarkkojs: I did the absolute minimal to keep that script happy.

ci/jobs/build-guests.sh
jarkkojs: It's not that different to Rust where you can do bunch of stuff with `{}`. I'd recommend to take these into use :-) The logic is not complex once you know them

jarkkojs:
I'll revisit the remaining issues on Monday. Took 4 weeks in the end to get here but zero actual technical debt generated (maybe even a bit reduced in details). It somehow exploded from that once additional `ld` parameter I added to the original JSON (`--relocatable`).

jarkkojs:
Also I think Monday would be good time because then the response time any possible issues is fastest (next week I'll be available full 5 days).

ci/jobs/build-guests.sh
xermicus: Does this work fine with zsh too? Bash only features will make Mac userss.

jarkkojs: It's formally same as asking does this work on a Python interpreter too. Why do you find difficult to install bash? I use zsh too.

Zsh has the construct I described, if that was the question.

guest-programs/build-benchmarks.sh
jarkkojs: I fixed up this one.

ci/jobs/build-guests.sh
jarkkojs: For what is worth, I checked running `zsh polkavm/guest-programs/build-examples.sh`, so if you really want to go that route, it is possible. Zsh as a scripting language is by a large part superset of bash.

crates/polkavm-linker/src/program_from_elf.rs
jarkkojs: Since I have the new bug to make the fix already opened, I linked the suggestion there, and will resolve this.

jarkkojs:
I addressed the reported issues so I'll merge this (including #199).

PR #188, opened 2024-10-10
Add blob length to container

xermicus:
Add the final blob length as a LE u64 to the container, so it can be efficiently parsed while still keeping the blob itself opaque.

This would solve https://github.com/paritytech/revive/issues/69


crates/polkavm-common/src/program.rs
```diff
@@ -3806,33 +3812,46 @@ pub struct ProgramParts {
...
-        };
+        })
+    }
+
+    pub fn blob_length(blob: ArcBytes) -> Result<u64, ProgramParseError> {
+        let mut reader = Reader::from(&blob);
+        Self::try_header(&mut reader)?;
+
+        Ok(u64::from_le_bytes(
+            blob[BLOB_LEN_RANGE].try_into().expect("BLOB_LEN_RANGE is a slice of 8 bytes; qed"),
```
koute, line 3848: This ain't substrate and we don't do these kinds of `expect`s here. (:

In general since you're doing a `u64::from_le_bytes` you're already hardcoding the length of the field, so there's probably little point in having a constant `BLOB_LEN_RANGE` for it? That said, what you could do is to make a type alias for `u64`, and then instead of `BLOB_LEN_RANGE` you'd just do `offset..offset + core::mem::size_of::<$type>()` and `$type::from_le_bytes`.

crates/polkavm/src/tests.rs
```diff
@@ -2342,6 +2343,12 @@ fn test_basic_debug_info() {
     assert!(frame.path().unwrap().unwrap().ends_with("/guest-programs/test-blob/src/main.rs"));
 }
 
+#[test]
+fn blob_len_works() {
+    const DOOM: &[u8] = include_bytes!("../../../examples/doom/roms/doom.polkavm");
```
koute, line 2348: You can use `guest-programs/output/example-hello-world.polkavm` instead of DOOM. (Doesn't matter too much, but it will bloat the test binary slightly less)

crates/polkavm-common/src/program.rs
```diff
@@ -4809,6 +4835,10 @@ pub const SECTION_END_OF_FILE: u8 = 0;
 
 pub const BLOB_VERSION_V1_64: u8 = 0;
 pub const BLOB_VERSION_V1_32: u8 = 1;
+pub const BLOB_VERSION_V2_64: u8 = BLOB_VERSION_V1_64 + 2;
+pub const BLOB_VERSION_V2_32: u8 = BLOB_VERSION_V1_32 + 2;
+pub const BLOB_VERSION_CURRENT_64: u8 = BLOB_VERSION_V2_64;
+pub const BLOB_VERSION_CURRENT_32: u8 = BLOB_VERSION_V2_32;
```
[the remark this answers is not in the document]
xermicus: Right, can just leave it at V1

crates/polkavm-common/src/program.rs
```diff
@@ -3806,33 +3812,46 @@ pub struct ProgramParts {
...
 
-        let mut parts = ProgramParts {
+        Ok(ProgramParts {
             is_64_bit,
             ..ProgramParts::default()
-        };
+        })
+    }
+
+    pub fn blob_length(blob: ArcBytes) -> Result<u64, ProgramParseError> {
```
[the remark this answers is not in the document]
xermicus: It seemed like a sensible idea to have at least the magic and version checked, but doing no checks is fine for me too.

crates/polkavm/src/tests.rs
xermicus: Ah nice thanks

crates/polkavm-common/src/program.rs
xermicus: Aight :) Removed the expect

pgherveou:
Nice thanks for adding this @xermicus 🙏

pgherveou:
This needs to be moved outside of polkavm_linker since it is not nostd and we need this blob_length function for pallet-revive

crates/polkavm/src/tests.rs
```diff
@@ -10,12 +10,13 @@ use alloc::vec;
...
 
 use polkavm_common::abi::MemoryMapBuilder;
-use polkavm_common::program::Reg::*;
 use polkavm_common::program::{asm, DefaultInstructionSet};
+use polkavm_common::program::{BlobLen, Reg::*};
 use polkavm_common::utils::align_to_next_page_u32;
 use polkavm_common::writer::ProgramBlobBuilder;
 
 use paste::paste;
+use polkavm_linker::ProgramParts;
```
pgherveou, line 19: maybe just  use polkavm_common::program::ProgramParts;

xermicus: Done, thanks

athei (approved):
Thanks!

crates/polkavm-common/src/program.rs
```diff
@@ -3806,6 +3812,14 @@ pub struct ProgramParts {
 }
 
 impl ProgramParts {
+    pub fn blob_length(raw_blob: &[u8]) -> Option<BlobLen> {
```
koute, line 3815: Ultimately it doesn't really matter but this would probably a little bit more appropriate on `ProgramBlob`, considering this is a public API? `ProgramParts` is really mostly here for JAM to be able to build a `ProgramBlob` from a non `.polkavm` blob, and since this is explicitly checking for the length of a `.polkavm` blob it'd be better to add it to the struct which represents a `.polkavm` blob. (:

crates/polkavm-common/src/program.rs
```diff
@@ -3819,6 +3833,9 @@ impl ProgramParts {
         };
 
         let blob_version = reader.read_byte()?;
+
+        reader.read_slice(BLOB_LEN_SIZE)?;
```
koute, line 3837: Hm, maybe verify that the length is correct at the end of the parsing?

crates/polkavm-common/src/program.rs
xermicus: Makes sense to me, changed it

crates/polkavm-common/src/program.rs
xermicus: Yes I agree, I think if the length mismatches the blob is invalid anyways so there is no point in continue parsing it.

crates/polkavm/src/tests.rs
```diff
@@ -2342,6 +2342,12 @@ fn test_basic_debug_info() {
     assert!(frame.path().unwrap().unwrap().ends_with("/guest-programs/test-blob/src/main.rs"));
 }
 
+#[test]
+fn blob_len_works() {
+    const DOOM: &[u8] = include_bytes!("../../../guest-programs/output/example-hello-world.polkavm");
+    assert_eq!(Some(DOOM.len() as BlobLen), ProgramBlob::blob_length(DOOM));
```
koute, line 2348: You forgot to rename the variable, as it's still `DOOM`. :P

PR #189, opened 2024-10-10
fix: macos installation steps

sacha-l:
These steps worked for me on `master`. The command that was here didn't work for me.


koute:
Yes, the generic backend on master is currently broken, so the macos instructions weren't technically working.

I'm fine with merging this, but this should have a note that it will use the interpreter so it will be slow/unplayable.

PR #192, opened 2024-10-13
polkavm-linker: Rework linker code optimization pass to make it work on 64-bit

aman4150:


crates/polkavm-linker/src/dwarf.rs
```diff
@@ -573,14 +573,20 @@ fn try_fetch_relocation(
...
-        )));
+    let target = match relocation {
+        RelocationKind::Abs {
+            target,
+            size: RelocationSize::U64,
+        } => target,
+        RelocationKind::Abs {
+            target,
+            size: RelocationSize::U32,
+        } => target,
```
koute, line 584: This should probably check whether we're on rv32 or rv64 and only match the 32-bit relocation on 32-bit, and 64-bit relocation on 64-bit.

crates/polkavm-linker/src/dwarf.rs
```diff
@@ -2218,17 +2228,7 @@ where
...
-                    size: RelocationSize::U32,
-                } = relocation
-                else {
-                    return Err(ProgramFromElfError::other(format!(
-                        "failed to process DWARF: failed to fetch DW_AT_low_pc for a unit: unexpected relocation at {relocation_target}: {relocation:?}"
-                    )));
-                };
-
-                return Ok(Some(*target));
+                return Ok(Some(target));
```
koute, line 2231: Hmm... this is probably incorrect?

For now it's fine if DWARF doesn't 100% work (just use `--strip` when running the linker to disable DWARF processing), so I'd probably revert this change.

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3912,7 +3956,7 @@ impl BlockRegs {
                 if kind == AnyAnyKind::Add
                     && src1_value != RegValue::Constant(0)
                     && src2_value != RegValue::Constant(0)
-                    && (src1_value.bits_used() & src2_value.bits_used()) == 0
```
koute, line 3915: This optimization will still be incorrect on 64-bit.

On 64-bit what `Add` does is that it:
1) adds the registers together,
2) ignores the upper 32-bits of the result,
3) sign extends the 32-bit result to 64-bits.

So let's say you have the following two values added:

```
  0b10100000000000000000000000000000
+ 0b01000000000000000000000000000001
```

So on 32-bit this will give you:

```
  0b11100000000000000000000000000001
```

But on 64-bit you will get:

```
  0b11111111111111111111111111111111_11100000000000000000000000000001
```

(All of the 32-bit instructions on 64-bit work this way.)

So in this case if this optimization is triggered on 64-bit you will end up with a value that is not sign extended to full 32-bits (because there is actually no 32-bit `Or` instruction; we *do* currently have dedicated 32-bit `Or` and 64-bit `Or`, but Cyril was somewhat overzealous when adding those - in RISC-V there's only a 64-bit one, so most likely we will remove the 32-bit, *unless* we can synthesize a 32-bit `Or` instructions ourselves *and* it will provide a measurable performance improvements, but it's too early to be doing that without benchmarking).

Honestly, the easiest way to deal with this (for now) would probably be for optimizations like this to disable them on 64-bit, and we can bring them back later? (We don't have to completely migrate the whole of the optimizer in one step.)

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3743,31 +3782,35 @@ impl RegValue {
 
     fn bits_used(self) -> u64 {
         match self {
-            RegValue::InputReg(..) | RegValue::CodeAddress(..) | RegValue::DataAddress(..) => u64::from(u32::MAX),
-            RegValue::Constant(value) => u64::from(value as u32),
+            RegValue::InputReg(..) | RegValue::CodeAddress(..) | RegValue::DataAddress(..) => u64::MAX,
```
koute, line 3785: Hm, maybe it'd be better to add a `Bitness` to `InputReg` and other variants of this enum that don't have `bits_used`? I don't particularly like that you have to manually mask the bits used after calling `bits_used()`; the `bits_used()` should be default always return the correct value.

crates/polkavm-linker/src/dwarf.rs
aman4150: Yes, that's right, but we do not have enough information here to do that, and passing rv64 flag seems like an overkill...considering we check for invalid relocations way before we reach here.

Currently the code throws `found an export with an unexpected relocation at ...` if it sees a 64-bit relocation in a 32-bit elf. I propose that we keep this change as-is.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: I agree, we can do it. But note that we will still be doing a mask where we are currently masking, reason being that now bits_used is 64-bit and for 32-bit elf pass we will end up calculating final bits_used mask beyond 32-bits.

I propose that we add bitness in `BlockRegs`, and mask the values when setting registers...

crates/polkavm-linker/src/program_from_elf.rs
aman4150: alright, I will disable this for now.

crates/polkavm-linker/src/dwarf.rs
aman4150: Ack

crates/polkavm-linker/src/dwarf.rs
aman4150: the check is actually made outside of the dwarf file, I will add a check/assert in dwarf code file for this.

koute (approved):

PR #195, opened 2024-10-15
Minor decoder refactoring

aman4150:
```
    polkavm-linker: Add a DecoderConfig struct
    
    We plan to extend decoder to include Zbb instruction set, while keeping
    the backward compatibility of the decoder in mind, we should add a flag
    to turn on this instruction set.
    
    Since we already have a flag for 64-bit, it will soon become ugly to keep
    passing a bunch of parameters to the decoder. Let's create a DecoderConfig
    to bundle the flags for the decoder.
    
    This patch only aims to refactor code, and no functional change is intended.
    
    Signed-off-by: Aman <aman@parity.io>
```

```
    polkavm-linker: Make decode_compressed(...) a private function
    
    ...and while at it, only do compressed decoding if that is intended or
    required, and do not fallback to normal decoding even if instruction
    byte code suggests compressed instruction.
    
    Signed-off-by: Aman <aman@parity.io>
```

### DIFF
--- a/crates/polkavm-linker/src/riscv.rs
+++ b/crates/polkavm-linker/src/riscv.rs
@@ -37,6 +37,26 @@ pub enum Reg {
     T6,
 }
 
+pub struct DecoderConfig {
+    pub(crate) rv64: bool,
+}
+
+impl DecoderConfig {
+    pub fn new() -> Self {
+        DecoderConfig { rv64: false }
+    }
+
+    #[cfg(test)]
+    pub fn new_64bit() -> Self {
+        DecoderConfig { rv64: true }
+    }
+
+    pub fn set_rv64(&mut self, rv64: bool) -> &mut Self {
+        self.rv64 = rv64;
+        self
+    }
+}
+
 impl Reg {
     pub const NAMES: &'static [&'static str] = &[
         "zero", "ra", "sp", "gp", "tp", "t0", "t1", "t2", "s0", "s1", "a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7", "s2", "s3", "s4",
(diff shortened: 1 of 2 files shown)


crates/polkavm-linker/src/riscv.rs
```diff
@@ -37,6 +37,26 @@ pub enum Reg {
     T6,
 }
 
+pub struct DecoderConfig {
+    pub(crate) rv64: bool,
+}
```
koute, line 42: Hm, do you see this having any other config knobs besides 64-bit? (As I can't think of any.)

(To be clear, it's fine if you want to refactor the `bool` to something else, I don't really care either way; just asking.)

aman4150: As I noted in the commit message, I do plan to add Zbb knob. Ideally moving forward, when we have this thing in production we do want every change behind feature flag, to always support backward compatibility (that is throw error for instruction set which is not supported by a particular version of linker?).

koute: For the VM, sure, but for the linker I don't think we really need to? I think we can just keep it simple and unconditionally always decode them.

And even *if* we would want the linker to reject programs with Zbb instruction (because e.g. that support is not finished, or something) then probably putting the check at the final codegen stage when emitting the blob would be better? You'll have to modify the `BasicInst` enum to add the `Zbb` instructions, so putting a check at the decoder won't magically make it so that the rest of the code doesn't have to deal with them, and even if "make sure no Zbb instruction gets emitted" is something we want the most appropriate thing would be to either make it a compile-time flag (so there is no change whatsoever that those instructions can exist), or also check for this as late as possible (since *technically* some of these instructions could get injected not at the decode stage, but also by the optimizer, etc., and putting the check before the final codegen would protect against that).

crates/polkavm-linker/src/riscv.rs
```diff
@@ -37,6 +37,26 @@ pub enum Reg {
...
+
+impl DecoderConfig {
+    pub fn new() -> Self {
```
koute, line 45: Naming-wise I think it'd be better for this to be `new_32bit` or something like that.

(If you named this based on what's in my WIP 64-bit branch: I had `new` and `new_64bit` there not because I want it that way, but because I wanted to get that code working as soon as possible, and the `new` method was already used in a bunch of places so I would have to fix them up if I renamed them. :P But for my final PR I would rename it.)

koute (approved):

PR #204, opened 2024-10-29
Add support for RISC-V Zbb (Bit manipulation) extension

aman4150:


crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3860,6 +4050,32 @@ impl OperationKind {
...
+            Self::MaximumUnsigned => (lhs as u64).max(rhs as u64) as i64,
+            Self::Minimum => lhs.min(rhs),
+            Self::MinimumUnsigned => (lhs as u64).min(rhs as u64) as i64,
+            Self::RotateLeft => {
+                if rv64 {
+                    cast((lhs as u64).rotate_left(rhs as u32)).to_signed()
+                } else {
+                    cast((lhs as u32).rotate_left(rhs as u32) as i32).to_i64_sign_extend()
+                }
+            },
```
koute, line 4069: Okay, so for this instruction specifically I would want to see a separate variant (instead of the `rv64` flag) like it is for e.g. addition where you have a dedicated "32 bit + sign extend" variant. Otherwise anywhere else you will want to do any optimizations on these operations you'll have to remember to have this extra `if` and if you forget it'll be a bug; having this as an extra variant will remove this potential footgutn

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3860,6 +4050,32 @@ impl OperationKind {
...
+            Self::RotateLeftWord => cast((lhs as u32).rotate_left(rhs as u32) as i32).to_i64_sign_extend(),
```
koute, line 4070: Can you rename this one to `RotateLeft32AndSignExtend`?

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3980,6 +4196,30 @@ impl OperationKind {
...
+            (O::MaximumUnsigned, C(0), rhs) => rhs,
+            (O::MaximumUnsigned, lhs, C(0)) => lhs,
+
+            (O::MinimumUnsigned, C(0), _rhs) => C(0),
+            (O::MinimumUnsigned, _lhs, C(0)) => C(0),
+
+            (O::RotateLeft, lhs, C(0)) => lhs,
+            (O::RotateLeft, C(0), _) => C(0),
+
+            (O::RotateLeftWord, lhs, C(0)) => lhs,
```
koute, line 4214: Isn't this incorrect? :P

(This will truncate the `lhs` and sign extend it to 64-bit instead of being a nop, no? See, that's the reason I was explicit about `32`/`64` and `32AndSignExtend` everywhere. :P)

aman4150: > This will truncate the lhs and sign extend it to 64-bit instead of being a nop
And that is right thing to do.

RotateLeftWord is not a 32-bit only instruction, this is what the spec says:
```
let rs1 = EXTZ(X(rs1)[31..0])
let shamt = X(rs2)[4..0];
let result = (rs1 << shamt) | (rs1 >> (32 - shamt));
X(rd) = EXTS(result[31..0]);
```

crates/polkavm-common/src/program.rs
```diff
@@ -1560,6 +1565,18 @@ define_opcodes! {
...
+
+        [I_64, I_32] and_inverted                             = 162,
+        [I_64, I_32] or_inverted                              = 137,
+        [I_64, I_32] xnor                                     = 138,
+        [I_64, I_32] maximum                                  = 145,
+        [I_64, I_32] maximum_unsigned                         = 146,
+        [I_64, I_32] minimum                                  = 147,
+        [I_64, I_32] minimum_unsigned                         = 148,
+        [I_64, I_32] rotate_left                              = 152,
+        [I_64, I_32] rotate_left_word                         = 153,
```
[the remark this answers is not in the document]
aman4150: No, rotate_left and rotate_left_word are available on both 32-bit and 64-bit.

The `rotate_left_word` is same as `rotate_left` on 32-bit, but on 64-bit it yield a sign-extended result.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: Sure, we could, but then it conflicts with our convention of using 32 and 64 in the operation kind name.

I suppose, I could rename is to `RotateLeftWordSignExtend` .

crates/polkavm-linker/src/program_from_elf.rs
aman4150: Hm, there is no reason for actually doing sign extension here, I guess I simply copied the statement from above and replaced it with `to_i64_sign_extend`.

On 32-bit, it will get truncated again, so sign-extending or not, it's the same story.

aman4150: I suppose, I can still split the this operation type, but this does seem a bit redundant. Because anywhere we don't extend the result type to match the return value, code would not compile.

I will however, replace `to_i64_sign_extend` with a simply cast with zero-extend op.

crates/polkavm-assembler/src/amd64.rs
```diff
@@ -1282,6 +1295,21 @@ pub mod inst {
             None,
             (fmt.write_fmt(core::format_args!("mov {}, {}", self.1.name_from(self.0), self.2.name_from(self.0)))),
 
+        movsx_8_to_64(Reg, Reg) =>
```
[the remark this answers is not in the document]
aman4150: Sure, not sure how I missed these.

crates/polkavm-common/src/program.rs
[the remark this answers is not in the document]
koute: > The `rotate_left_word` is same as `rotate_left` on 32-bit, but on 64-bit it yield a sign-extended result.

We shouldn't have duplicate opcodes doing the same thing, considering our limited opcode space.

Okay, so:

* `rotate_left_word` -> `rotate_left_32`
* `rotate_left` -> `rotate_left_64`
* `rotate_left_32` should be enabled on both 32-bit and 64-bit
* `rotate_left_64` should be enabled only on 64-bit
* `rotate_left_32` should not sign extend on 32-bit, and sign extend on 64-bit

That's how all of the rest of the instructions are currently implemented.

crates/polkavm-linker/src/program_from_elf.rs
koute: Well, you can just do it the same way the rest of the ops are implemented (e.g. look at addition or shifts). :P

crates/polkavm-linker/src/program_from_elf.rs
koute: The convention so far in the linker is:
- `64` in the name - used only on 64-bit
- `32AndSignExtend` - used only on 64-bit, originally a 32-bit instruction
- `32` - used only on 32-bit

Fundamentally the 32-bit rotate left is a different operation on 32-bit than on 64-bit. On 32-bit it doesn't do any sign extension, and on 64-bit it truncates the 64-bit value to 32-bit, does the operation, and sign extends back to 64-bit. So even though in RISC-V these are the same instruction the semantics are very different between 32-bit and 64-bit targets, so they should have a different variant in the linker.

crates/polkavm/src/compiler/amd64.rs
```diff
@@ -1527,6 +1782,72 @@ where
...
+    }
+
+    #[inline(always)]
+    pub fn count_set_bits_word(&mut self, d: RawReg, s: RawReg) {
+        self.push(popcnt(RegSize::R32, conv_reg(d), conv_reg(s)))
+    }
+
+    #[inline(always)]
+    pub fn sign_extend_byte(&mut self, d: RawReg, s: RawReg) {
+        self.push(movsx_8_to_64(conv_reg(d), conv_reg(s)))
```
koute, line 1817: Add `RegSize` argument to `movsx_*` instructions and on 32-bit only sign extend it to 32-bit.

In the recompiler we *don't* have the convention that 32-bit values are always sign extended to full 64-bit, and the 32-bit version of movsx is one byte shorter since it doesn't need the REX prefix.

crates/polkavm/src/compiler/amd64.rs
```diff
@@ -1527,6 +1782,72 @@ where
...
+    #[inline(always)]
+    pub fn or_combine_byte(&mut self, d: RawReg, s: RawReg) {
+        let reg_size = self.reg_size();
+        let d = conv_reg(d);
+        let s = conv_reg(s);
+        let or_combine_label = self.or_combine_label;
+
+        self.push(mov(reg_size, TMP_REG, s));
+        self.call_to_label(or_combine_label);
+        self.push(mov(reg_size, d, TMP_REG));
```
koute, line 1839: Use `self.asm.reserve`. Feel free to do `asm.push(call_label32(label))` directly; the `call_to_label` helper is kinda useless and I just forgot to delete it :P

crates/polkavm-common/src/program.rs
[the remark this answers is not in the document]
aman4150: Will fix this.

crates/polkavm-assembler/src/amd64.rs
```diff
@@ -1517,6 +1545,59 @@ pub mod inst {
...
+            None,
+            (fmt.write_fmt(core::format_args!("rol {}, cl", self.1.display(Size::from(self.0))))),
+
+        ror_cl(RegSize, RegMem) =>
+            Inst::new(0xd3).rex_64b_if(matches!(self.0, RegSize::R64)).regmem(self.1).modrm_opext(0b001).encode(),
+            None,
+            (fmt.write_fmt(core::format_args!("ror {}, cl", self.1.display(Size::from(self.0))))),
+
+        // https://www.felixcloutier.com/x86/popcnt
+        popcnt(RegSize, Reg, RegMem) =>
```
koute, line 1559: Still missing more instructions not added to the tests... :P

crates/polkavm-common/src/program.rs
```diff
@@ -1576,6 +1593,17 @@ define_opcodes! {
...
+        [I_64]       count_leading_zero_bits_word             = 140,
+        [I_64, I_32] count_trailing_zero_bits                 = 141,
+        [I_64]       count_trailing_zero_bits_word            = 142,
+        [I_64, I_32] count_set_bits                           = 143,
+        [I_64]       count_set_bits_word                      = 144,
+        [I_64, I_32] sign_extend_byte                         = 149,
+        [I_64, I_32] sign_extend_half_word                    = 150,
+        [I_64, I_32] zero_extend_half_word                    = 151,
+        [I_64, I_32] or_combine_byte                          = 156,
+        [I_64, I_32] reverse_byte                             = 157,
```
koute, line 1606: Can we adjust the naming of these `_word` instructions like we did for rotate so that they're all consistent?

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -1942,6 +2077,10 @@ where
                 };
             }
 
+            if resolve_simple_zero_register_usage(kind, dst, src1, src2, &mut emit) {
+                return Ok(());
```
koute, line 2081: This should not return. We should emit a NOP here.

For every RISC-V instruction at least one internal instruction should be emitted, or else we'll get weird bugs in corner cases (ask me how I learned this the hard way.... :P); I forgot to add an assertion which enforces this (you can add one if you want).

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -1702,6 +1716,94 @@ fn emit_minmax(
     }));
 }
 
+fn resolve_simple_zero_register_usage(
```
koute, line 1719: Is there any particular reason why you added this? AFAIK if all of this is properly handled in `apply`/`apply_const` then we shouldn't need this at all, no?

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -3980,6 +4221,34 @@ impl OperationKind {
...
+            (O::RotateLeft64, C(0), _) => C(0),
+
+            (O::RotateRight32, lhs, C(0)) => lhs,
+            (O::RotateRight32, C(0), _) => C(0),
+
+            (O::RotateRight32AndSignExtend, C(0), _) => C(0),
+
+            (O::RotateRight64, lhs, C(0)) => lhs,
+            (O::RotateRight64, C(0), _) => C(0),
+
```
koute, line 4251: This looks a little funny; you've added comments for the first two, then you gave up on adding comments, but still nicely aligned the third one, and then you gave up on alignment for the rest. :P

Can you add comments for all of them and also nicely align them?

(Also, there are two arms which use `_lhs` and `_rhs` instead of `_`; doesn't really matter in the grand scheme of things, but that does trigger my OCD a little bit. :P)

The arms look correct to me now though. :+1:

crates/polkavm-linker/src/program_from_elf.rs
aman4150: My understanding is that apply or apply_const can only return a register move or load immediate instruction. In this subroutine we are handling different kind of instruction replacement, such as where we need 1s complement or sign extend the register value.

I agree that in the case when optimizer is enabled, we may not need most of these (if not all), but this subroutine makes sure we do not throw an error when running with linker optimizer disabled.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: I am really interested in that story now :P

crates/polkavm-common/src/program.rs
aman4150: Sure

crates/polkavm-linker/src/program_from_elf.rs
amaneureka: Don't we run `simplify_instruction` only when `config.optimize` is true?

I'll keep it here for now, we can discuss it in detail tomorrow.

koute: We do as part of the optimizer. But we can also selectively run it only for those instructions which use the zero register when optimizations are disabled, just to get rid of the zero register usage.

crates/polkavm-linker/src/program_from_elf.rs
koute: Right, if this is guaranteed to insert an instruction then adding a nop is, of course, not necessary, but in that case we really need an assert somewhere to make sure that's the case. Just from looking at this prototype it's not obvious that it actually emits an instruction, and looking at the definition of the `resolve_simple_zero_register_usage` it's not obvious that it's necessary to emit one, so this is bad, and with nothing to enforce this it's just a footgun waiting to blow up. :P

aman4150: Ack, yes, it is not enforced. I've added NOP for now.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: Fixed in the latest iteration.

koute (approved):

PR #213, opened 2024-11-10
 Protect VmCtx and ID_COUNTER with RwLock

jarkkojs:
```
    Protect VmCtx and ID_COUNTER with RwLock
    
    AtomicU64 causes compilation issues when used as a subcrate of another
    crate when the toolchain configurations conflict. The problem was observed
    with ID_COUNTER when Polkadot SDK was compiled for riscv32emac, which could
    not be found from core.
    
    The documentation also states that:
    
    "This type is only available on platforms that support atomic loads and
    stores of u64."
    
    Address this by protecting ID_COUNTER with `spin::RwLock` and `VmCtx` with
    `std::sync::RwLock`.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -459,6 +459,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "01cda141df6706de531b6c46c3a33ecca755538219bd484262fa09410c13539c"
 
 [[package]]
+name = "lock_api"
+version = "0.4.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07af8b9cdd281b7915f413fa73f29ebd5d55d0d3f0155584dade1ff18cea1b17"
+dependencies = [
+ "autocfg",
+ "scopeguard",
+]
+
+[[package]]
 name = "log"
 version = "0.4.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -584,6 +594,7 @@ dependencies = [
  "log",
  "polkavm-assembler",
  "proptest",
+ "spin",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


jarkkojs:
Noteworthy argument is also that a register state itself is an atomic state. Using atomic for different registers does not guarantee this any better than u64.

jarkkojs:
Turned this into draft to give some peace to finalize my idea/PoC. I'll undraft it when I seek for comments.

jarkkojs:
This is from discussion where the conclusion was that kernel will never use Rust's atomic type, as it BTW does not use C's atomic types:

https://lwn.net/ml/linux-kernel/CAHk-=whY5A=S=bLwCFL=043DoR0TTgSDUmfPDx2rXhkk3KANPQ@mail.gmail.com/

Linus is absolutely correct in what he is saying here. Generally language provided atomic is types are unreliable and should be avoided when there is no actual need. By heavily deploying atomic like it is deployed like now in the code base without compiler we make the code only more sensitive to the compiler bugs. And also incompatibilities as documentation specifically says not to support `AtomicU64` on 32-bit architectures.

Also in most cases where atomic is used and where I migrated to plain u64 it has no value. E.g. register state is defined as the atomic snapshot of all registers.

To summarize that using more coarse locking we make the code more robust and compatible despite some minor differences in the environment.

Thus, I ended up that better granularity is to protect `VmCtx` and `ID_COUNTER`. This should make the code a factor more reliable too: you can change this with the lock and be sure that all other threads see only the final state.

In the crates I used `spin`, which I've previous used in https://github.com/enarx/enarx and also Google recommends this crate for bare metal, which also PolkaVM is (it is software-define bare-mental or this is how I like to think of it):

https://google.github.io/comprehensive-rust/bare-metal/useful-crates/spin.html

jarkkojs:
From Google's recommended bare metal I'd also recommend to check out:

- https://github.com/rcore-os/buddy_system_allocator
- https://github.com/Lokathor/tinyvec
- https://github.com/google/zerocopy/

E.g. we could in future consider do we need our own allocator or would widely use allocator crate do the job (out-of-scope but it is just one thing I've spotted where we could tighten up).

jarkkojs:
One thing noticed: convert Box to Arc:

1. Box: drop + free
2. Arc: drop

jarkkojs:
> One thing noticed: convert Box to Arc:
> 
>     1. Box: drop + free
> 
>     2. Arc: drop

BOTH suck as I can just do this:

```rust
        let vmctx = RwLock::new(unsafe { &mut *vmctx_mmap.as_mut_ptr().cast::<VmCtx>() });
```

And then assign that to the struct :-) Does the job...

PR #216, opened 2024-11-12
chore: fix broken link in `riscv`

zerosnacks:
Updates https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#instruction-aliases to https://github.com/riscv-non-isa/riscv-asm-manual/blob/main/src/asm-manual.adoc#instruction-aliases


koute:
Thanks.

PR #219, opened 2024-11-14
Embed target JSON in the linker crate

koute:


crates/polkavm-linker/src/lib.rs
```diff
@@ -11,3 +13,43 @@ mod utils;
...
+
+pub static TARGET_JSON_32_BIT: &str = include_str!("../riscv32emac-unknown-none-polkavm.json");
+pub static TARGET_JSON_64_BIT: &str = include_str!("../riscv64emac-unknown-none-polkavm.json");
+
+fn target_json_path_impl(json_name: &str, json_contents: &str) -> Result<PathBuf, String> {
+    let version = env!("CARGO_PKG_VERSION");
+    let cache_path = dirs::cache_dir()
+        .or_else(|| std::env::current_dir().ok())
+        .unwrap_or_else(|| PathBuf::from("./"))
+        .join(".polkavm-linker")
```
nazar-pc, line 25: If under cache, why hidden directory?

PR #230, opened 2024-11-25
Update opcodes

subotic:


subotic:
@koute RfR. I think that an approval is required for the CI jobs to run, since this is my first PR in this organization.

subotic:
@koute I've recompiled the example (on MacOS), and now the tests pass locally. You would need to again approve the workflow.

aman4150:
I've approved the workflow on @koute's behalf.

subotic:
@aman4150 Should I update the Rust version as it is suggested in the error:
```
Caused by:
  package `cargo-platform v0.1.9` cannot be built because it requires rustc 1.78 or newer, while the currently active rustc version is 1.75.0
```

subotic:
@aman4150 Could you maybe approve the workflow again?

koute:
Why the MSRV bump? What exactly was failing?

koute:
So the issue here is that `cargo install cargo-fuzz` is failing on the CI. The way to fix it is to just use the updated toolchain to install cargo-fuzz itself, and not to bump the toolchain version for the whole project.

subotic:
> So the issue here is that `cargo install cargo-fuzz` is failing on the CI. The way to fix it is to just use the updated toolchain to install cargo-fuzz itself, and not to bump the toolchain version for the whole project.

Sorry about that. I have updated the `rust-toolchain.toml` inside the `fuzz` directory, but I don't know if this is enough, i.e. if it is additionally necessary to run `cargo install cargo-fuzz` inside the `fuzz` directory. Could you please approve the workflow again? Unfortunately, the CI is not running on my fork although I have enabled it, so I cannot test it without bothering you.

koute:
@subotic Yeah, they way you did it is not going to work.

This need to be fixed [here](https://github.com/paritytech/polkavm/blob/master/.github/workflows/rust.yml#L100), and an easy fix might be to `rustup toolchain install $toolchain` first and then prepend `rustup run $toolchain` to the `cargo install cargo-fuzz` command.

Fundamentally the problem here is that the `cargo install cargo-fuzz` on the CI is just using whatever toolchain is already installed, which is probably the one that's in the toplevel `rust-toolchain.toml` in the repo, and since I haven't pinned the `cargo-fuzz` version (which we should also probably do) it just pulls the newest version, and the newest versions seems to require a newer toolchain now, hence your PR's failing.

subotic:
@koute Thanks for the explanation. I've added the changes you have suggested. Can you please approve the workflow?

.github/workflows/rust.yml
```diff
@@ -1,107 +1,111 @@
...
+    fuzz:
+        runs-on: ubuntu-22.04
+        steps:
+            -   uses: actions/checkout@v4
+            -   name: Install cargo fuzz
+                run: |
+                    rustup toolchain install $TOOLCHAIN
+                    rustup run $TOOLCHAIN cargo install cargo-fuzz
+                env:
+                    TOOLCHAIN: 1.83.0
```
koute, line 109: Just hardcode it without the env.

.github/workflows/rust.yml
```diff
@@ -1,107 +1,111 @@
...
-    - uses: actions/checkout@v4
-    - name: Install cargo fuzz
-      run: cargo install cargo-fuzz
-    - name: Fuzz
-      run: ./ci/jobs/fuzz.sh
+    build-and-test-linux:
+        runs-on: ubuntu-22.04
+        steps:
+            -   uses: actions/checkout@v4
+            -   name: Enable unprivileged userfaultfd
```
koute, line 17: Can you revert the formatting changes?

fuzz/rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly-2024-07-10"
+channel = "1.83.0"
```
koute, line 2: This can now be reverted.

The fuzz crate itself needs a nightly toolchain, so it won't work with a stable one.

.github/workflows/rust.yml
subotic: Sure, I have extended `.editorconfig`.

subotic:
@koute Could you please approve the workflow? And sorry for the noise. This is my first PR in this organisation.

.github/workflows/rust.yml
```diff
@@ -13,95 +13,97 @@ jobs:
...
     runs-on: windows-latest
     steps:
-    - uses: actions/checkout@v4
-    - name: Build and test
-      shell: bash
-      run: ./ci/jobs/build-and-test.sh
+      - uses: actions/checkout@v4
+      - name: Build and test
+        shell: bash
+        run: ./ci/jobs/build-and-test.sh
```
koute, line 45: This is still reformatted. (:

subotic:
@koute thanks for merging.

PR #238, opened 2024-12-06
spectool: add 64-bit test programs (GP-0.5.3)

subotic:
- added test programs for 64-bit instructions as per GP-0.5.3
- extended "polkavm-common/assembler" to support 64-bit instructions
- fixed some unit tests. Need pointers on how to fix them correctly.


subotic:
@koute Could you maybe take a look if I'm on the right track?

crates/polkavm-common/src/program.rs
```diff
@@ -1435,7 +1435,7 @@ define_opcodes! {
         [I_64, I_32] load_u16                                 = 54,
         [I_64, I_32] load_i16                                 = 55,
         [I_64, I_32] load_i32                                 = 57,
-        [I_64]       load_u32                                 = 56,
+        [I_64, I_32] load_u32                                 = 56,
```
koute, line 1438: Revert this change; for 32-bit there is only `load_i32` (it would be equivalent to `load_u32` anyway, since the registers are only 32-bit there)

koute:
> @koute Could you maybe take a look if I'm on the right track?

Yes, you're on the right track. (:

crates/polkavm-common/src/program.rs
subotic: Thanks, I don't really understand what the macro is doing. I was just looking at the other instructions and it seem to me, as it would be missing here.

These parameters, do they signify the PVM type, i.e., PVM running in 32-bit vs 64-bit?

subotic:
@koute Also, is it possible to authorise me to run the workflows? I don't want to bother you all the time.

subotic:
@koute Is it correct, that `load_imm` given an 4 byte immediate like `0xdeadbeef`, the post condition of the register should be `0x00000000deadbeef` and not `0xffffffffdeadbeef`?

subotic:
@koute how to discern between `load_imm` and load_imm64` if the disassembly for both is the same `{d} = 0x{a:x}`? Can/should I add to  `load_imm` or `load_imm64` something like `i64 a0 = 10`? Or am I misunderstanding `load_imm64` and the value of the immediate needs to be always larger then `i32`?

koute:
> Also, is it possible to authorise me to run the workflows? I don't want to bother you all the time.

Not entirely sure how to do that; I've added you as a read-only contributor - hopefully that will make the tests auto run.

> Is it correct, that load_imm given an 4 byte immediate like 0xdeadbeef, the post condition of the register should be 0x00000000deadbeef and not 0xffffffffdeadbeef? Or should the logic be that it is always sign extended? 

If it's a negative number (i.e. the most significant bit is set) then yes, it should be sign extended (but the VM takes care of that automatically).

> how to discern between load_imm and load_imm64 if the disassembly for both is the same {d} = 0x{a:x}? Can/should I add to load_imm64 something like i64 a0 = 10? Or am I misunderstanding load_imm64 and the value of the immediate needs to be always larger then i32?

If a value can be encoded with a `load_imm` then `load_imm` should always be used. If the value cannot be encoded with `load_imm` then only then `load_imm_64` should be emitted.

(And we could also have an override to force `load_imm_64` for testing, e.g. maybe `{reg} = i64 {value}`?)

subotic:
Great pointers thank you very much. This clears and answers a lot.

subotic:
@koute could you maybe take a look again? I had to change some unit tests, which should probably be done differently, but for this I would need your input.

crates/polkavm-common/src/assembler.rs
```diff
@@ -427,10 +443,25 @@ pub fn assemble(code: &str) -> Result<Vec<u8>, String> {
...
+                    }
+                }
+
                 if let Some(imm) = parse_imm(rhs) {
                     emit_and_continue!(Instruction::load_imm(dst.into(), imm as u32));
                 }
 
+                if let Some(imm64) = parse_imm64(rhs) {
+                    emit_and_continue!(Instruction::load_imm64(dst.into(), imm64 as u64));
+                }
```
koute, line 463: This is partially my fault because of how the load_imm instructions are currently printed out (I need to change it, or you could change it if you want), but here's how it should work:

* `a0 = $value` in the assembly/disassembly code should *always* result it *exactly* the $value being loaded into the register (what you see is what you get)
* by default if the immediate can be represented with `load_imm` then it always should use `load_imm`, and only should fall back to `load_imm64` when it cannot
* `i64` prefix should always force `load_imm64`
* `load_imm` can be used to load immediates between `0` and `0x7fffffff` (inclusive), and also any immediates where all upper 33 bits are set (because it sign extends)

So, for example:

* `a0 = 0xffffffff` should result in `0xffffffff` being assigned, so this must use `load_imm64` (because `load_imm` will sign extend the value from 32-bits to 64-bits)
* `a0 = 0xffffffff87654321` should result in `0xffffffff87654321` being assigned and should use `load_imm` by default (because all of the upper bits are `1`s so `load_imm` with argument `0x87654321` can be used)
* `a0 = -2` should result in `0xfffffffffffffffe` being assigned and should use `load_imm` by default (because all of the upper bits are `1`s so `load_imm` with argument `0xfffffffe` can be used)

tools/spectool/src/main.rs
```diff
@@ -118,9 +118,16 @@ fn parse_pre_post(line: &str, output: &mut PrePost) {
...
-        let rhs = cast(rhs).to_u64();
+        let rhs = polkavm_common::utils::parse_imm(rhs)
+            .map(|i| {
+                if i < 0 {
+                    cast(cast(i).to_i64_sign_extend()).to_unsigned()
+                } else {
+                    cast(cast(i).to_unsigned()).to_u64()
+                }
+            })
+            .or_else(|| polkavm_common::utils::parse_imm64(rhs).map(|i| cast(i).to_unsigned()))
```
koute, line 129: This should work similar to how I described `load_imm` should work (so what value you see is what you get if it's in hex, and if it's like e.g. `-4` then it should be immediately parsed as i64)

tools/spectool/spec/src/inst_add_32.txt
```diff
@@ -0,0 +1,9 @@
+pre: a0 = 1
+pre: a1 = 2
```
koute, line 2: For these 32-bit addition tests it would be nice to have:

* Both input value are larger than 32-bit, to show that the 32-bit version effectively truncates them
* Have two versions of the test -  one which results in sign extension (the 31th bit of the result is one) and one which results in no sign extension (the 31th bit of the result is zero), to show the sign extension mechanic

koute (commented):
Thanks for the PR!

tools/spectool/src/main.rs
```diff
@@ -135,8 +142,15 @@ fn main_generate() {
     let root = Path::new(env!("CARGO_MANIFEST_DIR")).join("spec");
     let mut found_errors = false;
 
-    for entry in std::fs::read_dir(root.join("src")).unwrap() {
-        let path = entry.unwrap().path();
+    let mut entries: Vec<std::fs::DirEntry> = std::fs::read_dir(root.join("src"))
+        .unwrap()
+        .filter_map(|entry| entry.ok())
```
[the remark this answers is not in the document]
subotic: Done

tools/spectool/spec/src/inst_add_32.txt
subotic: I've added two more programs. Can you please take a look and see if this is what you where thinking?

crates/polkavm-common/src/assembler.rs
subotic: @koute Ok, I think I got it now. Can you take a look at the new `parse_immediate` function and see if it correctly captures the described behaviour?

tools/spectool/src/main.rs
subotic: now using the new `parse_immediate` function.

crates/polkavm-common/src/utils.rs
```diff
@@ -284,6 +284,67 @@ pub fn parse_imm(text: &str) -> Option<i32> {
...
+            Ok(signed) => signed as u64,
+            Err(_) => return None,
+        }
+    };
+
+    if force_imm64 {
+        return Some(ParsedImmediate::U64(value));
+    }
+
+    if (value <= 0x7FFF_FFFF) || (value & 0xFFFF_FFFF_0000_0000 == 0xFFFF_FFFF_0000_0000) {
```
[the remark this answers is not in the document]
subotic: Thanks. Your approach is much simpler and more readable in code.

tools/spectool/spec/src/inst_shift_arithmetic_right_imm_alt_64.txt
```diff
@@ -4,5 +4,5 @@ pre: gas = 10000
 pub @main:
     a2 = 0x80000075 >>a a1
 
-post: a2 = 0xf000000e
+post: a2 = 0xfffffffff000000e
```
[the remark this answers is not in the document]
subotic: Thanks. I know where the problem lies.

koute (approved):

Issue #209, opened 2024-11-01
Remove non-DOOM ELF binaries from `test-data`


koute:
Since we're using an upstream toolchain now we don't have to keep `bench-pinky` and `test-blob` ELF binaries in `test-data` anymore; we should instead trigger their generation as needed from the tests themselves.

jarkkojs:
Which binaries test-data file map exactly (32/64 versions)?

jarkkojs:
**PoC:** https://github.com/paritytech/polkavm/commit/713cb7adee8392de2f1b4762a55be246f041014a

Step 1:

```
error: couldn't read crates/polkavm/src/../../../guest-programs/target/bench-pinky_32.elf: No such file or directory (os error 2)
    --> crates/polkavm/src/tests.rs:1502:25
     |
1502 |         get_blob(false, include_bytes!("../../../guest-programs/target/bench-pinky_32.elf"))
     |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)

error: couldn't read crates/polkavm/src/../../../guest-programs/target/bench-pinky_64.elf: No such file or directory (os error 2)
    --> crates/polkavm/src/tests.rs:1504:25
     |
1504 |         get_blob(false, include_bytes!("../../../guest-programs/target/bench-pinky_64.elf"))
     |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)

error: couldn't read crates/polkavm/src/../../../guest-programs/target/test-blob_32.elf: No such file or directory (os error 2)
    --> crates/polkavm/src/tests.rs:1943:33
     |
1943 | const TEST_BLOB_32_ELF: &[u8] = include_bytes!("../../../guest-programs/target/test-blob_32.elf");
     |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)

[...]

PR #240, opened 2024-12-15
polkavm-test-data

jarkkojs:
Closes: #209


jarkkojs:
I don't know how to fix build-and-test-windows.

athei:
The windows runner doesn't seem to use a rustup managed cargo. So they are not able to download and use the toolchain specified in the `rustup-toolchain.toml`. So you might have to make sure that rustup is used on windows.

crates/polkavm-test-data/build.rs
```diff
@@ -0,0 +1,64 @@
...
+    let remap = std::format!(
+        "--remap-path-prefix={}= --remap-path-prefix={}=",
+        polkavm.into_os_string().into_string().unwrap(),
+        std::env!("HOME"),
+    );
+
+    let mut cmd = Command::new("cargo");
+
+    cmd.current_dir(project_path)
+        .env_clear()
```
koute, line 17: Why the `env_clear`?

crates/polkavm-test-data/build.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::process::Command;
+
+fn build(project: &str, profile: &str, target: &str) -> Result<(), String> {
+    let polkavm = Path::new(std::env!("CARGO_MANIFEST_DIR")).join("../..");
+    let project_path = polkavm.join("guest-programs").join(project).into_os_string().into_string().unwrap();
+
+    let remap = std::format!(
+        "--remap-path-prefix={}= --remap-path-prefix={}=",
+        polkavm.into_os_string().into_string().unwrap(),
+        std::env!("HOME"),
```
koute, line 11: The first remap was previously from `guest-programs` to an empty string (currently it is from the repo root to an empty string), and the second remap was previously to `~` instead of to an empty string.

koute:
> I don't know how to fix build-and-test-windows.

This is weird; it's clearly using rustup because `cargo test --all` triggers a toolchain install:

```
>> cargo test (debug)
info: syncing channel updates for '1.75.0-x86_64-pc-windows-msvc'
info: latest update on 2023-12-28, rust version 1.75.0 (82e1608df 2023-12-21)
```

but when it's ran from inside of the `build.rs` script it ignores the `guest-programs/rust-toolchain.toml` for some reason?

I'm not sure how to fix it; you'll have to fiddle with it. Maybe clearing the environment variables screws up with it?

crates/polkavm-test-data/build.rs
```diff
@@ -0,0 +1,64 @@
...
+    );
+
+    let mut cmd = Command::new("cargo");
```
athei, line 14: Not sure if it resolves the windows problem but it seems best practice to do this in build scripts:

```suggestion
    let mut cmd = Command::new(std::env!("CARGO"));
```

athei:
> This is weird; it's clearly using rustup because `cargo test --all` triggers a toolchain install:

Oops didn't see that and jumped to conclusions. Rustup seems to installed and is picking up the toolchain file in the repo root.

> but when it's ran from inside of the `build.rs` it ignores the `guest-programs/rust-toolchain.toml` for some reason?

Yes it is weird. Maybe some platform differences regarding pathes?

> I'm not sure how to fix it; you'll have to fiddle with it. Maybe clearing the environment variables screws it up?

We do the clearing in the other build scripts in polkadot-sdk, too.  But I don't think it is ever tested on Windows. Yeah maybe removing the clearing and work from there. Maybe additional or different variables need to be forwarded on Windows.

crates/polkavm-test-data/build.rs
jarkkojs: Not sure. I need refresh on Monday. It's a strong possibility that it is for no good reason ;-)

crates/polkavm-test-data/build.rs
jarkkojs: thanks, agreed

crates/polkavm-test-data/build.rs
jarkkojs: OK, I'll revisit also on Monday.

crates/polkavm-test-data/build.rs
jarkkojs: The history for this is that I derived the code from substrate/frame/revive/fixtures/build.rs.

I guess the argument here (and fixtures) is that by doing this we remove any side-effects of environment variables set by parent i.e. always starting from the same state.

jarkkojs: I recall @athei did the code I used as basis so maybe he has a comment for this...

crates/polkavm-test-data/build.rs
jarkkojs: Right I simply made a mistake here, thanks!

crates/polkavm-test-data/build.rs
athei: Yeah it makes sense since we don't want other variables meant for other targets to influence this build.

athei:
At least now it is failing everywhere? :D

jarkkojs:
> At least now it is failing everywhere? :D

Working on it... temporary glitch ;---)

jarkkojs:
@koute, @athei, I'll go a few diff's now that I've fixed the stuff I know how to fix!

Diff 1:

```
diff --git a/crates/polkavm-test-data/build.rs b/crates/polkavm-test-data/build.rs
index 2def01d..4971115 100644
--- a/crates/polkavm-test-data/build.rs
+++ b/crates/polkavm-test-data/build.rs
@@ -21,7 +21,6 @@ fn build(project: &str, profile: &str, target: &str) -> Result<(), String> {
     let mut cmd = Command::new("cargo");
 
     cmd.current_dir(project_path)
-        .env_clear()
         .env("PATH", std::env!("PATH"))
         .env("RUSTFLAGS", rust_flags)
         .env("RUSTUP_HOME", std::env!("RUSTUP_HOME"))
```

Result:

```
~/work/github.com/paritytech/polkavm/crates/polkavm-test-data polkavm-test-data*
$ cargo b    
warning: /home/jarkko/work/github.com/paritytech/polkavm/Cargo.toml: unused manifest key: workspace.lints.rust.unexpected_cfgs.check-cfg
   Compiling polkavm-test-data v0.18.0 (/home/jarkko/work/github.com/paritytech/polkavm/crates/polkavm-test-data)
error: failed to run custom build command for `polkavm-test-data v0.18.0 (/home/jarkko/work/github.com/paritytech/polkavm/crates/polkavm-test-data)`

Caused by:
  process didn't exit successfully: `/home/jarkko/work/github.com/paritytech/polkavm/target/debug/build/polkavm-test-data-0ea5ae9020752356/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=OUT_DIR

  --- stderr
  Error: "error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel\nSee https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.\n"
```

Diff 2:

```
diff --git a/crates/polkavm-test-data/build.rs b/crates/polkavm-test-data/build.rs
index 2def01d..b637903 100644
--- a/crates/polkavm-test-data/build.rs
+++ b/crates/polkavm-test-data/build.rs
@@ -18,10 +18,9 @@ fn build(project: &str, profile: &str, target: &str) -> Result<(), String> {
 
     // Note that using `std::env!("CARGO")` fails the compilation since the
     // compilation requires nightly.
-    let mut cmd = Command::new("cargo");
+    let mut cmd = Command::new(std::env!("CARGO"));
 
     cmd.current_dir(project_path)
-        .env_clear()
         .env("PATH", std::env!("PATH"))
         .env("RUSTFLAGS", rust_flags)
         .env("RUSTUP_HOME", std::env!("RUSTUP_HOME"))
```

Result:

```
~/work/github.com/paritytech/polkavm/crates/polkavm-test-data polkavm-test-data* 24s
$ cargo b    
warning: /home/jarkko/work/github.com/paritytech/polkavm/Cargo.toml: unused manifest key: workspace.lints.rust.unexpected_cfgs.check-cfg
   Compiling polkavm-test-data v0.18.0 (/home/jarkko/work/github.com/paritytech/polkavm/crates/polkavm-test-data)
error: failed to run custom build command for `polkavm-test-data v0.18.0 (/home/jarkko/work/github.com/paritytech/polkavm/crates/polkavm-test-data)`

Caused by:
  process didn't exit successfully: `/home/jarkko/work/github.com/paritytech/polkavm/target/debug/build/polkavm-test-data-0ea5ae9020752356/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=OUT_DIR

  --- stderr
  Error: "error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel\nSee https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.\n"
```

Diff 3:

```
diff --git a/crates/polkavm-test-data/build.rs b/crates/polkavm-test-data/build.rs
index 2def01d..a901351 100644
--- a/crates/polkavm-test-data/build.rs
+++ b/crates/polkavm-test-data/build.rs
@@ -18,7 +18,7 @@ fn build(project: &str, profile: &str, target: &str) -> Result<(), String> {
 
     // Note that using `std::env!("CARGO")` fails the compilation since the
     // compilation requires nightly.
-    let mut cmd = Command::new("cargo");
+    let mut cmd = Command::new(std::env!("CARGO"));
 
     cmd.current_dir(project_path)
         .env_clear()
```

Result:

```
~/work/github.com/paritytech/polkavm/crates/polkavm-test-data polkavm-test-data* 19s
$ cargo b    
warning: /home/jarkko/work/github.com/paritytech/polkavm/Cargo.toml: unused manifest key: workspace.lints.rust.unexpected_cfgs.check-cfg
   Compiling polkavm-test-data v0.18.0 (/home/jarkko/work/github.com/paritytech/polkavm/crates/polkavm-test-data)
error: failed to run custom build command for `polkavm-test-data v0.18.0 (/home/jarkko/work/github.com/paritytech/polkavm/crates/polkavm-test-data)`

Caused by:
  process didn't exit successfully: `/home/jarkko/work/github.com/paritytech/polkavm/target/debug/build/polkavm-test-data-0ea5ae9020752356/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=OUT_DIR

  --- stderr
  Error: "error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel\nSee https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.\n"
```

crates/polkavm-test-data/build.rs
jarkkojs: See my remarks below.

athei:
> How should I move forward?

Debug this further. Probably reproducing in a Windows VM. Looking at the code it seems like it should work so I am out of ideas by just looking at it.

jarkkojs:
> > How should I move forward?
> 
> Debug this further. Probably reproducing in a Windows VM. Looking at the code nothing obvious comes to mind for me.

I'm working on a Windows VM. It is worth of investment as Windows could break also in future because POSIX does not always align that well in that environment.

jarkkojs:
I spent a lot of time on setting up a Windows environment that is usable. Now I have an image that logins automatically and boots to the terminal, and I can reach the host via ssh. Also environment has been setup.

This what happens:

polkavm-test-data

jarkkojs:
I think it was useful spend couple of days to figure out effective way to use for Windows. After this I can reproduce bugs in no time. It was an investment ;-)

jarkkojs:
> I spent a lot of time on setting up a Windows environment that is usable. Now I have an image that logins automatically and boots to the terminal, and I can reach the host via ssh. Also environment has been setup.
> 
> This what happens:

In the previous, I failed to test correctly: I should have run `ci/jobs/build-and-test.sh` in a bash prompt, not `cargo t` in a PowerShell prompt.

This what I get now:

```
>> cargo test (debug)
warning: C:\Users\jarkko\work\github.com\paritytech\polkavm\Cargo.toml: unused manifest key: workspace.lints.rust.unexpected_cfgs.check-cfg
warning: unused imports: `ISA64_V1_NoSbrk`, `build_static_dispatch_table`
 --> crates\polkavm\src\api.rs:9:5
  |
9 |     build_static_dispatch_table, FrameKind, ISA32_V1_NoSbrk, ISA64_V1_NoSbrk, Imports, InstructionSet, Instructions,...
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^                              ^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused variable: `exports`
   --> crates\polkavm\src\api.rs:410:13
    |
410 |         let exports = {
    |             ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_exports`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: method `context` is never used
  --> crates\polkavm\src\error.rs:74:19
   |
62 | impl Error {
   | ---------- method in this implementation
...
74 |     pub(crate) fn context(self, message: impl core::fmt::Display) -> Self {
   |                   ^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field `engine_state` is never read
   --> crates\polkavm\src\api.rs:229:5
    |
228 | pub(crate) struct ModulePrivate {
    |                   ------------- field in this struct
229 |     engine_state: Option<Arc<EngineState>>,
    |     ^^^^^^^^^^^^

warning: methods `compiled_module` and `address_to_page` are never used
   --> crates\polkavm\src\api.rs:285:19
    |
263 | impl Module {
    | ----------- methods in this implementation
...
285 |     pub(crate) fn compiled_module(&self) -> &CompiledModuleKind {
    |                   ^^^^^^^^^^^^^^^
...
343 |     pub(crate) fn address_to_page(&self, address: u32) -> u32 {
    |                   ^^^^^^^^^^^^^^^

warning: variants `Disjoint`, `None`, `One`, and `Two` are never constructed
  --> crates\polkavm\src\page_set.rs:21:5
   |
20 | enum SubResult {
   |      --------- variants in this enum
21 |     Disjoint,
   |     ^^^^^^^^
22 |     None,
   |     ^^^^
23 |     One(Interval),
   |     ^^^
24 |     Two(Interval, Interval),
   |     ^^^
   |
   = note: `SubResult` has derived impls for the traits `Debug` and `Clone`, but these are intentionally ignored during dead code analysis

warning: associated items `merge` and `subtract` are never used
  --> crates\polkavm\src\page_set.rs:28:8
   |
27 | impl Interval {
   | ------------- associated items in this implementation
28 |     fn merge(self, rhs: Interval) -> Option<Interval> {
   |        ^^^^^
...
41 |     fn subtract(existing: Interval, to_remove: Interval) -> SubResult {
   |        ^^^^^^^^

warning: associated items `new`, `insert`, `contains`, `is_whole_region_empty`, `remove`, and `clear` are never used
   --> crates\polkavm\src\page_set.rs:147:12
    |
146 | impl PageSet {
    | ------------ associated items in this implementation
147 |     pub fn new() -> Self {
    |            ^^^
...
153 |     pub fn insert(&mut self, (insert_min, insert_max): (u32, u32)) {
    |            ^^^^^^
...
188 |     pub fn contains(&self, (min, max): (u32, u32)) -> bool {
    |            ^^^^^^^^
...
203 |     pub fn is_whole_region_empty(&self, (min, max): (u32, u32)) -> bool {
    |            ^^^^^^^^^^^^^^^^^^^^^
...
218 |     pub fn remove(&mut self, (removed_min, removed_max): (u32, u32)) {
    |            ^^^^^^
...
272 |     pub fn clear(&mut self) {
    |            ^^^^^

warning: associated items `new_reusing_memory`, `len`, and `clear` are never used
  --> crates\polkavm\src\utils.rs:63:12
   |
50 | / impl<T> FlatMap<T>
51 | | where
52 | |     T: Copy,
   | |____________- associated items in this implementation
...
63 |       pub fn new_reusing_memory(mut memory: Self, capacity: u32) -> Self {
   |              ^^^^^^^^^^^^^^^^^^
...
75 |       pub fn len(&self) -> u32 {
   |              ^^^
...
85 |       pub fn clear(&mut self) {
   |              ^^^^^

warning: method `lock` is never used
  --> crates\polkavm\src\mutex_std.rs:12:12
   |
5  | impl<T> Mutex<T> {
   | ---------------- method in this implementation
...
12 |     pub fn lock(&self) -> std::sync::MutexGuard<T> {
   |            ^^^^

warning: `polkavm` (lib) generated 10 warnings (run `cargo fix --lib -p polkavm` to apply 2 suggestions)
   Compiling polkavm-test-data v0.18.0 (C:\Users\jarkko\work\github.com\paritytech\polkavm\crates\polkavm-test-data)
   Compiling fnv v1.0.7
error: failed to run custom build command for `polkavm-test-data v0.18.0 (C:\Users\jarkko\work\github.com\paritytech\polkavm\crates\polkavm-test-data)`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_TEST_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `C:\Users\jarkko\work\github.com\paritytech\polkavm\target\debug\build\polkavm-test-data-6cd164aff147f2c5\build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-env-changed=OUT_DIR

  --- stderr
  Error: "error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel\nSee https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.\n"
warning: build failed, waiting for other jobs to finish...
```
This looks like the same issue as the one that pops up CI, doesn't it?

@athei, @koute: I give up now for this week but check if you spot anything interesting (and put a note).

athei:
Yeah it looks like the same issue. The toolchain file is ignored and hence it is using the default toolchain which doesn't support the unstable features we are using.

jarkkojs:
This gets more interesting:
```
jarkko@DESKTOP-8GF9ULF MINGW64 ~/work/github.com/paritytech/polkavm/guest-programs (polkavm-test-data)
$ rustup show
Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\jarkko\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc (default)
nightly-2024-11-01-x86_64-pc-windows-msvc
1.75.0-x86_64-pc-windows-msvc

active toolchain
----------------

nightly-2024-11-01-x86_64-pc-windows-msvc (overridden by 'C:\Users\jarkko\work\github.com\paritytech\polkavm\guest-programs\rust-toolchain.toml')
rustc 1.84.0-nightly (a0d98ff0e 2024-10-31)
```

I.e. totally what you'd expect o_O

jarkkojs:
I got a bit further. Now I'm bumping to:

```
  Error: "error: linker `link.exe` not found\n  |\n  = note: program not found\n\nnote: the msvc targets depend on the msvc linker but `link.exe` was not found\n\nnote: please ensure that Visual Studio 2017 or later, or Build Tools for Visual Studio were installed with the Visual C++ option.\n\nnote: VS Code is a different product, and is not sufficient.\n\nerror: could not compile `proc-macro2` (build script) due to 1 previous error\nerror: could not compile `compiler_builtins` (build script) due to 1 previous error\n"
warning: build failed, waiting for other jobs to finish...
```

I've tried bunch of thing like:

1. stable-x86_64-pc-windows-msvc -> stable-x86_64-pc-windows-gnu
2. Install vs_BuildTools.exe in addition Visual Studion Community edition.

Also:

```
jarkko@DESKTOP-8GF9ULF MINGW64 ~/work/github.com/paritytech/polkavm (polkavm-test-data)
$ link
link: missing operand
Try 'link --help' for more information.
```

Windows is always an interesting journey to lobotomy ;-)

jarkkojs:
@athei, @koute New problem in Windows. Since our build script is in bash, it will conflict with link.exe because bash environment also has /usr/bin.

Possible routes forward:

1. Implement build-and-test.cmd
2. Switch to link-lld for the sake of resolving name conflict and also speeding up a bit [1].

What is it going to be?

[1] https://dsincl12.medium.com/faster-rust-builds-on-windows-7a7662c16f9

jarkkojs:
@koute, @athei: I found a way to workaround the issue: `rustup run nightly cargo` (pushed).

jarkkojs:
@koute, @athei 

Oops in a bit bad state but I cannot update it because:

```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Somehow my access is now denied.

jarkkojs:
> @koute, @athei: I found a way to workaround the issue: `rustup run nightly cargo` (pushed).

In practice not so great...

jarkkojs:
@athei, @koute: Looks familiar: https://github.com/rust-lang/rust/issues/37868

athei:
This reads as if this is only about the path of the executable itself. Not about files relative to he executable.

Can you please avoid force pushing open PRs. It is hard to track changes over time otherwise. Making the work of reviewers harder.

jarkkojs:
> This reads as if this is only about the path of the executable itself. Not about files relative to the executable.
> 
> Can you please avoid force pushing open PRs. It is hard to track changes over time otherwise. Making the work of reviewers harder.

Here this is essential because important is to execute cargo in the correct directory. I.e. when cargo is launched it needs to be in guest-programs so that rust-toolchain.toml is taken into account.

jarkkojs:
Converted to draft.

athei:
I would understand this argument if your commit message would not just be a single "polkavm-test-data". If you were maintaining a nice patch set with well described and broken down commits I would understand that you want to rebase to maintain that nice history.

But just rewriting a single commit that contains all changes without a proper description is just the worst of both worlds. We neither can track what we already have reviewed nor do we have a nice commit history.

> About force-pushing. Well I can turn it into draft but polkavm github does not do automatic squashes.

But you can select to squash and merge or is this disabled?  But a draft is acceptable for force pushing. I will not look at drafts then, though.

jarkkojs:
@athei with this the extra issue with managing the PR comes from bad development environment in Windows, and have not been able to establish a shared directory. So only I need to pro-actively push stuff in order to in turn compile it in Windows. It is what it is ...

jarkkojs:
If I add `+nightly` to cargo:

```
error: failed to run custom build command for `polkavm-test-data v0.19.0 (C:\Users\jarkko\work\github.com\paritytech\polkavm\crates\polkavm-test-data)`

Caused by:
  process didn't exit successfully: `C:\Users\jarkko\work\github.com\paritytech\polkavm\target\debug\build\polkavm-test-data-255e6fbec63893f3\build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-env-changed=OUT_DIR
  cargo::error=error: no such command: `+nightly`

        Cargo does not handle `+toolchain` directives.
        Did you mean to invoke `cargo` through `rustup` instead?
```

If I try `rustup run nightly cargo build`:

```
error: failed to run custom build command for `polkavm-test-data v0.19.0 (C:\Users\jarkko\work\github.com\paritytech\polkavm\crates\polkavm-test-data)`

Caused by:
  process didn't exit successfully: `C:\Users\jarkko\work\github.com\paritytech\polkavm\target\debug\build\polkavm-test-data-255e6fbec63893f3\build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-env-changed=OUT_DIR
  cargo::error=error: "C:\\Users\\jarkko\\.rustup\\toolchains\\1.75.0-x86_64-pc-windows-msvc\\lib\\rustlib\\src\\rust\\library\\Cargo.lock" does not exist, unable to build with the standard library, try:
          rustup component add rust-src --toolchain nightly-x86_64-pc-windows-msvc


  --- stderr
  Error: ()
warning: build failed, waiting for other jobs to finish...
```
(rust-src is installed)

jarkkojs:
@athei Just a proof that this does work:

```
PS C:\Users\jarkko\work\github.com\paritytech\polkavm\guest-programs> cargo build -q --profile no-lto --bin test-blob -p test-blob --target ..\crates\polkavm-linker\riscv64emac-unknown-none-polkavm.json -Zunstable-options -Zbuild-std="core,alloc"
PS C:\Users\jarkko\work\github.com\paritytech\polkavm\guest-programs> fd -I -H -t x test-blob
target\riscv64emac-unknown-none-polkavm\no-lto\.fingerprint\test-blob-c22761cea0639592\bin-test-blob
target\riscv64emac-unknown-none-polkavm\no-lto\.fingerprint\test-blob-c22761cea0639592\dep-bin-test-blob
target\riscv64emac-unknown-none-polkavm\no-lto\test-blob
```

jarkkojs:
@athei I found a solution, it is unclean but the point here is the cmd-indirection that seems to solve the issue. I can make it take "project" and "profile" as parameter so it will match build_unix. After that all is cleaned up and this commit is split to sane peaces etc. we can land this.

jarkkojs:
Hmm... OK so no luck today I guess... I think that this cmd indirection might possibly be still worth of investigating further.

jarkkojs:
@athei I'll keep this as single commit up until Windows issue is rooted and fixed, and the split this to a nice set of patches with description. Before that it is not really worth of doing.

jarkkojs:
Hooray: 



I was able to share my Git root to Windows VM by using these instructions. Should help with demand for force push:

https://www.heiko-sieger.info/sharing-files-between-the-linux-host-and-a-windows-vm-using-virtiofs/

jarkkojs:
I was able to install OpenSSH Server AND share the folder so less force pushing:

Screenshot From 2025-01-18 13-10-55

They are in the same folder. Pretty neat :-)

jarkkojs:
Without this Linux build will act the same (will complain about -Z flags):

```
    if !cfg!(target_os = "windows") {
        cmd.env_clear()
            .env("PATH", std::env::var("PATH").unwrap())
            .env("RUSTUP_HOME", std::env::var("RUSTUP_HOME").unwrap());
    }
```

So better to have this initial environment setup branched and find the right recipe for Windows.

jarkkojs:
@athei, @koute, possible success I got into brand new error:

```
error: failed to run custom build command for `polkavm-test-data v0.19.0 (D:\a\polkavm\polkavm\crates\polkavm-test-data)`

Caused by:
  process didn't exit successfully: `D:\a\polkavm\polkavm\target\debug\build\polkavm-test-data-38f55663277395bd\build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-env-changed=OUT_DIR
  cargo::error=info: syncing channel updates for 'nightly-2024-11-01-x86_64-pc-windows-msvc'
  error: could not download file from 'https://static.rust-lang.org/dist/2024-11-01/channel-rust-nightly.toml.sha256' to 'C:\Users\runneradmin\.rustup\tmp\p0cb_8xh58tanwbo_file': failed to make network request: error sending request for url (https://static.rust-lang.org/dist/2024-11-01/channel-rust-nightly.toml.sha256): client error (Connect): dns error: A non-recoverable error occurred during a database lookup. (os error 11003): A non-recoverable error occurred during a database lookup. (os error 11003)

warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

I removed `RUSTUP_HOME` setting from initial environment setup to get here.

jarkkojs:
Better to take a break with this one, nothing comes to mind ATM to try out with Windows build. Revisiting when new idea pop up.

jarkkojs:
Some new remarks.

I wrote a script yesterday (Sunday), which captures WMI system event, and matches process names per keyword when they are created. I can use that to capture environment variables from terminal and build.rs runs of cargo. I capture both cargo and rustup process spawns.

The script dumps environment variables to a text file.

Then as postmortem analysis I look at disjoint sets of these runs.

Probably env_clear() is a bit brutal action to take for any possible OS so that I'll take away completely.

jarkkojs:
Just putting reminder here. To make WinDbg work:

1. Disable Secure Boot
2. `bcdedit -debug on`

Then you can attach WinDbg to NT kernel: 

image

PR #248, opened 2025-01-07
Fix non-linux compilation warnings.

tomusdrw:
I've started adding `#[cfg(target_os = "linux")]` but for some cases it starts a cascade of related code not being used, so I think it's safer to go with conditionally allowing dead code.

Let me know what's your preference.


koute:
Hmmm... well, I'm not sure whether gating those to Linux is the way we want to go, due to a few reasons:

- it's not strictly correct (e.g. on Aarch64 Linux systems you'll also get extra warnings)
- we will want to support the recompiler on other OSes in the future too (And in fact the recompiler used to work on macOS too in the past, before I had to refactor a ton of code and didn't have the time to get the generic sandbox updated too.)
- and it seems kinda painful to maintain

So I think what we could do:
- in cases where the code in question is, not sure how to call it, "general purpose", we could unconditionally allow dead_code (for example, `Mutex::lock` should never give out a warning even when unused, because the whole point of a mutex is locking, so it doesn't make sense for it to not have this method)
- for compiler-specific code wrap it in `if_compiler_is_supported! {` (which is the proper way to gate code); unfortunately this is sometimes somewhat awkward to do (could be made easier with a procedural macro, but I refuse to bloat up the compile times for everyone just so that it's a little bit more convenient for me)
- Run clippy on the CI twice, but use a target which doesn't have a recompiler the second time to make sure the warnings don't come back

tomusdrw:
Thanks for the suggestions! I've left a few `allow(dead_code)` where I left it's general purpose. The rest was converted to use `if_compiler_is_supported`.
I've also added a clippy run on macos to prevent new warnings.

.github/workflows/rust.yml
```diff
@@ -36,6 +36,16 @@ jobs:
...
     - name: Build and test (macOS-only)
       run: ./ci/jobs/build-and-test-macos.sh
+  clippy-macos:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install clippy (base toolchain)
+      run: rustup component add clippy
+    - name: Install clippy (zygote toolchain)
+      run: cd crates/polkavm-zygote && rustup component add clippy
```
koute, line 46: Zygote is (and will always be) Linux-only so for macOS it should be removed. (I guess you'll need to modify the `clippy.sh` script to only run clippy on zygote on Linux; you can copy-paste the detection from `run-all-tests.sh`)

tomusdrw:
Apologies for the borked merge. I run the CI on my fork: https://github.com/FluffyLabs/polkavm/pull/1 and it passes, so it should be ready for re-review.

koute (approved):

koute:
Thanks!

PR #249, opened 2025-01-09
polkavm: Replace orc.b with an instruction sequence in the linker.

aman4150:
And get rid orc.b emulation code from the interpreter and the recompiler.

We are making this change because orc.b is rarely used and a very complicated instruction in itself, we don't need to support or upstream such a complicated instruction to Polkavm. Therefore let's emulate the instruction within the linker.

In addition, we are adding a warn print for the linker user to make them aware that orc.b is being replaced with an instruction sequence, and any perf benefit that the user is expecting would not be visible.

### DIFF
--- a/crates/polkavm-linker/src/program_from_elf.rs
+++ b/crates/polkavm-linker/src/program_from_elf.rs
@@ -1811,6 +1811,73 @@ fn resolve_simple_zero_register_usage(
     false
 }
 
+fn emit_or_combine_byte(
+    location: SectionTarget,
+    dst: Reg,
+    src: Reg,
+    rv64: bool,
+    mut emit: impl FnMut(InstExt<SectionTarget, SectionTarget>),
+) {
+    let op_reg = dst;
+    let cmp_reg = Reg::E1;
+    let tmp_reg = Reg::E2;
+    let mask_reg = if dst != src { src } else { Reg::E3 };
+    let range = if rv64 { 0..64 } else { 0..32 };
+
+    log::warn!("Emulating orc.b at {:?} with an instruction sequence", location);
+
+    if dst != src {
+        emit(InstExt::Basic(BasicInst::MoveReg { dst, src }));
+    }
+
+    // Loop:
+    // mov tmp, op
+    // shl mask, 8
+    // or tmp, mask
+    // test op, mask
+    // cmov.neq op, tmp
+
+    for iter in range.step_by(8) {
+        emit(InstExt::Basic(BasicInst::MoveReg { dst: tmp_reg, src: op_reg }));
+
+        if iter == 0 {
+            emit(InstExt::Basic(BasicInst::LoadImmediate { dst: mask_reg, imm: 0xff }));
+        } else {
+            emit(InstExt::Basic(BasicInst::AnyAny {
+                kind: if rv64 {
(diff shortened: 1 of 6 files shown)


crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -1811,6 +1811,73 @@ fn resolve_simple_zero_register_usage(
...
+    rv64: bool,
+    mut emit: impl FnMut(InstExt<SectionTarget, SectionTarget>),
+) {
+    let op_reg = dst;
+    let cmp_reg = Reg::E1;
+    let tmp_reg = Reg::E2;
+    let mask_reg = if dst != src { src } else { Reg::E3 };
+    let range = if rv64 { 0..64 } else { 0..32 };
+
+    log::warn!("Emulating orc.b at {:?} with an instruction sequence", location);
```
koute, line 1827: Should probably be a `debug`

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -2005,18 +2072,29 @@ where
             };
 
             use crate::riscv::RegKind as K;
+
+            //
+            // if OrCombineByte then emit function call
+            //
+
+            if kind == K::OrCombineByte {
```
koute, line 2080: You can move the body of this `if` into the `match` below and avoid the `unreachable!`

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -2005,18 +2072,29 @@ where
...
+                emit_or_combine_byte(current_location, dst, src, rv64, &mut emit);
+                emit(InstExt::nop());
```
koute, line 2082: Why emit the extra nop?

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -2005,18 +2072,29 @@ where
             };
 
             use crate::riscv::RegKind as K;
+
+            //
+            // if OrCombineByte then emit function call
```
koute, line 2077: Comment seems kinda superfluous (:

crates/polkavm-linker/src/program_from_elf.rs
aman4150: This is intentional. @athei recommended that we should emit a warning if orc.b is replaced by the linker. This allows a linker user to be well informed about the performance impact.

crates/polkavm-linker/src/program_from_elf.rs
aman4150: will do

crates/polkavm-linker/src/program_from_elf.rs
aman4150: Something I forget to remove, that's it, will remove it.

crates/polkavm-linker/src/program_from_elf.rs
athei: We assume that `orc.b` is emitted by hand tuned assembly, right? So it is probably rare and surprising when it is replaced by a slower sequence? So yes, I think a warning makes sense here IMHO.

koute (approved):

PR #252, opened 2025-01-20
fuzz: Add a linker fuzzer

aman4150:
We add a fuzzer for the linker, that uses RISC-V instruction encoding logic to create a sequence of valid instructions. Which then is compiled into a ELF binary and passed to the linker.

Expectation here is that the fuzzer would always generate a valid ELF binary with valid instructions, and the linker should be able to link it without any issues.

Currently the ELF binary is limited to one section (.text) and no symbols. We may want to extend that in the future.

Note: I saw some crashes that the fuzzer has reported and I am working on investigating those issues.

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -3,6 +3,18 @@
 version = 3
 
 [[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
 name = "arbitrary"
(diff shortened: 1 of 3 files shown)


jarkkojs:
This is the PR that I get notifications for that lead to 404 error every week for some reason...

subotic:
@jarkkojs Sorry to hear that, but no idea what is happening. I didn't work on it for some time now. @aman4150 took over, so I assume that my PR can be deleted.

subotic:
@jarkkojs Sorry to hear that, but no idea what is happening. I didn't work on it for some time now. @aman4150 took over, so I assume that my PR can be deleted.

jarkkojs:
> @jarkkojs Sorry to hear that, but no idea what is happening. I didn't work on it for some time now. @aman4150 took over, so I assume that my PR can be deleted.

No need for apologies, I was just wondering what it is :-) They did not appear for some time but I got notification again last week. Not Github expert so thus just plain asking...

koute (approved):

PR #256, opened 2025-01-23
Alter immediate parsing behaviour.

tomusdrw:
I think the `parse_immediate` function introduced in #238 is not handling correctly values from `0x7fff_ffff...0xffff_ffff`.

For instance:
```
r7 = r7 + 0xffffffff
```
fails to assemble, since the argument is `U64` and for `ADD_IMM_64` a 32-bit immediate is rather expected.
Note this particular code is produced, if you start off with `r7 = r7 - 1`, assemble that code and then disassemble it.

I don't see a good reason why negative `i32` values shouldn't be allowed to be provided as hex equivalents, however I obviously had to modify assertions in the tests, so curious if this is the right approach.

I've re-generated the test vectors based on this PR and you can find the results here:
https://github.com/FluffyLabs/jamtestvectors/pull/11

AFAICT, it's mostly changes from `LOAD_IMM_64` to `LOAD_IMM`, since the values are now interpreted as `U32`. Also tested running these vectors on my PVM impl and it passes cases from both PRs.


koute:
> `r7 = r7 + 0xffffffff` fails to assemble

That's expected as it's not possible to have such an offset logically. When assembled this will execute `r7 = r7 + 0xffffffffffffffff`, and `0xffffffff` is not the same as `0xffffffffffffffff`. :P

> I don't see a good reason why negative i32 values shouldn't be allowed to be provided as hex equivalents

Because it's confusing. If you want to specify a negative value either write it in decimal, or write it fully in hex.

Note that other assemblers also work this way, for a good reason. For example:

```
.intel_syntax noprefix

add [rax + 0xffffffff], rbx
```

Try to assemble this with `gcc test.s` and you'll get:

```
test.s: Assembler messages:
test.s:3: Error: 0xffffffff out of range of signed 32bit displacement
```

tomusdrw:
Hmm, that's definitely out of my depth and I have absolutely no experience in this, so I'll definitely trust you on this one :)

However what confuses me even more is:
1. > If you want to specify a negative value either write it in decimal, or write it fully in hex.
   
     The only way to produce valid code seems to be using decimals, since I can't write full hex value of a negative number (u32 vs u64 requirement), but:

1. > and 0xffffffff is not the same as 0xffffffffffffffff

    Sometimes it is! :)
   ```
   r7 = 0xffffffffff // this sign-extends to -1
   r8 = i64 0xffffffffff // this does not sign-extend 
   ```

1. > Note that other assemblers also work this way, for a good reason.
    
    Should the result produced by disassambler be changed since it can't be assembled back?

I understand the assembler/disassembler is meant to be experimental, so consider this as merely adding some data points about perceived inconsistencies.

From what I can tell, from asm-writer perspective, it's "safer" to use decimals, since they have more predictable behavior (i.e. always match the expected argument size of the instruction (i32 vs i64)).

Should the disassembler be allowed to be configured to produce decimals instead of hex output then (I see a potentially relevant option [here](https://github.com/FluffyLabs/polkavm/blob/d2f4ccd79a6018a602e04cf4663e30d0ebf99388/crates/polkavm-disassembler/src/lib.rs#L70))?

tomusdrw:
Ah, apologies I've realised what you meant just now.
```
r7 = 0xffffffff
```
This only sign-extends because of my changes (and it should not).

```
r7 = r7 + i64 0xffffffffffffffff
```

This is the proper way to write `r7 = r7 - 1`, since it would assemble to `ADD_IMM_64`.

I'm curious about the diassembler then, should it be "fixed" by either producing `64-bit` equivalent instructions and sign-extending the hex values, producing (negative) decimals for small numbers or neither?

koute:
> From what I can tell, from asm-writer perspective, it's "safer" to use decimals, since they have more predictable behavior (i.e. always match the expected argument size of the instruction (i32 vs i64)).

Both should be equally safe and have exactly the same predictability. The whole point is that what you get in the assembly should be what logically happens, e.g. if you have `a0 = a1 + $number` then the given number should be what it's logically added, regardless of how it's encoded.

> Should the result produced by disassambler be changed since it can't be assembled back?

I don't have anything currently to keep them in sync (e.g. I don't have a fuzz test for this), but in general yes, whatever the disassembler spits out should be a valid input to the assembler and should result in the same bytecode. If this is not a case feel free to create an issue and report it as a bug.

> This is the proper way to write `r7 = r7 - 1`

The proper way to write this is either `r7 = r7 - 1` or `r7 = r7 + 0xffffffffffffffff`. Both should work and produce the same bytecode; if not then it's a bug.

> I'm curious about the diassembler then, should it be "fixed" by either producing `64-bit` equivalent instructions and sign-extending the hex values, producing (negative) decimals for small numbers or neither?

The disassembler should always show what logically happens, so any immediates it shows should be in their fully sign-extended form. Whether it produces decimals or hex numbers is a matter of taste/configuration.

tomusdrw:
Understood! Thank you for your patience and descriptive response. I'll take a look again at what disassembler spits out exactly and perhaps create a PR/issue to address it (🤞 it's within JAM competition rules compliance, but I'm willing to give it a shot).

Issue #247, opened 2025-01-07
Bug when accessing string compiled into binary


cmichi:
I believe I found a bug in one of the `polkavm` components. I've put the reproducer [here](https://github.com/use-ink/polkavm/tree/cmichi-reproduce-bug-jan-25).

The reproducer program tries to output a string that is already present in its own blob (in the `.data` section, I think). But instead of printing the intended string (`"Success"` in the reproducer) we get an artifact from another string present in the blob. The length of the printed string is correct, the output is just a different part of another string from the blob.

```bash
cd examples/program-for-bug/
cargo +nightly build --release --no-default-features --target ../../crates/polkavm-linker/riscv64emac-unknown-none-polkavm.json -Zbuild-std="core,alloc" 

cd ../trigger-bug/
INPUT=../program-for-bug/target/riscv64emac-unknown-none-polkavm/release/program-for-bug cargo run
```

The output will be along the lines of
```
Calling into the guest program:
print: "heap: the nul"
print: "stack: "
print: "the nul"
```

This is wrong, it should be 
```
Calling into the guest program:
print: "heap: Success"
print: "stack: Success"
```

The string `the nul` is present in the blob and part of a longer string. This can be checked via `strings $INPUT`. The length (`7`) corresponds to the length of `"Success"`.

The bug also occurs if I use a different Rust version or don't use an allocator. If I don't include an allocator, the fragment that is output is just different.

Compiling `program-for-bug` without `--release` seems to mitigate the bug at least for the heap variant:

```
Calling into the guest program:
print: "heap: Success"
thread 'main' panicked at examples/trigger-bug/src/main.rs:48:10:
called `Result::unwrap()` on an `Err` value: Trap
```

cmichi:
Minimal reproducer:

```rust
#![no_std]
#![no_main]

extern crate core;

use polkavm_derive::polkavm_export;
use core::fmt::Write;

pub enum Foo {
    One,
    CalleeTrapped,
    Unknown,
}

impl ::core::fmt::Debug for Foo {
    #[inline]
    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
        ::core::fmt::Formatter::write_str(
            f,
            match self {
                Foo::Success => "Success",
                Foo::CalleeTrapped => "CalleeTrapped",
                Foo::Unknown => "Unknown",
            },
        )
    }
}

struct Writer;
impl core::fmt::Write for Writer {
[...]

PR #260, opened 2025-01-26
polkavm-linker: Swap origin/target in ADD32/SUB32

jarkkojs:
Relocation for ADD32/SUB32 relocation pair has origin/target wrong order. Swap origin and target.


jarkkojs:
Is this due the fix (candidate) or something else:

```

failures:

---- tests::compiler_linux_doom_o3_dwarf2 stdout ----
thread 'tests::compiler_linux_doom_o3_dwarf2' panicked at crates/polkavm/src/tests.rs:1380:72:
called `Result::unwrap()` on an `Err` value: ProgramFromElfError(Other("failed to process DWARF: unexpected relocation at <section #18+208>: Offset { origin: <section #1+2588>, target: <section #1+0>, size: Generic(U32) }"))


failures:
    tests::compiler_linux_doom_o3_dwarf2
```

jarkkojs:
Closes: #247

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -7960,6 +7960,17 @@ where
...
+                relocations.insert(
+                    current_location,
+                    RelocationKind::Offset {
+                        origin: *target_1,
+                        target: *target_2,
+                        size: SizeRelocationSize::Generic(RelocationSize::U32),
+                    },
+                );
+                continue;
+            }
```
koute, line 7973: Logically this isn't correct, so it's most likely not a proper fix (as evidenced by the failing tests).

We have `RelocationKind::Offset { origin, target, ... }`, so in `origin` we should have the base from which the offset is calculated, and in `target` we should have the destination, so logically this should match a relocation that looks like: `offset = target - origin`. However what you're doing here is the other way around: `offset = origin - target`.

guest-programs/program-for-bug/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    let _ = write!(&mut m, "{:?}", Foo::Success);
+}
+
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    unsafe {
+        core::arch::asm!("unimp");
+        core::hint::unreachable_unchecked();
+    }
+}
```
koute, line 56: This isn't the proper way to regression test this.

There's no guarantee that the compiler will actually emit the code which triggers the issue (and the issue very much depends on how *exactly* the compiler will compile this), unless we freeze the compiler version and the flags we use, which we don't want to do.

The best way to do this would probably be something like this:
1. Compile the program.
2. Disassemble it.
3. Strip down the assembly to the bare minimum.
4. Add the test as an assembly source code and reassemble it to produce a binary. (It's fine to commit the binary to not require everyone to install RISC-V assembler, but the blob should be reproducible for those who want to rebuild it.)

This would also allow the removal of part of the code which are unrelated to the core issue (e.g. the `debug_message` call is unnecessary and should be removed, etc.)

jarkkojs: What about using global_asm!() for wrapping it up? It's fairly robust

koute: > What about using global_asm!() for wrapping it up? It's fairly robust

That would be also acceptable, if it can be made to work.

(The major point here is that you need very specific relocations to be emitted to reproduce this bug, so it's not just an issue of emitting the right assembly but also getting the relocations right; you need to force the relocations to be emitted and not be mangled by the linker, etc.; no idea if that's doable with `global_asm!`)

jarkkojs: The second reasonable alternative

1. "
2. "
3. Use https://docs.rs/cc/latest/cc/ to compile assembly in build.rs of `crates/polkavm`.

This is almost what you suggested but has the benefit that we could edit the assembly code if we ever want to. I'll just try which is the most lean option for us (I honestly don't know before trying them out).

crates/polkavm-linker/src/program_from_elf.rs
jarkkojs: OK, cool so it was the reason for DOOM failing? :-) I suppose so as the master does not CI fail. I would not have open this PR if I had noticed that locally (not actually sure if I did go through the CI script or did just `cargo t -p polkavm-test-data`).

It was just place-holder to mark the location where I should do a fixup, and was quite surprised of the result.

jarkkojs:
Converted to draft up until the first flush of issues has been fixed. Thank you for the reviews.

guest-programs/program-for-bug/src/main.rs
jarkkojs: Out of my sleeve guess is that actually the second option is at least the one I try out first ;-) The first is great for bootstrapping code in some cases but probably not as convenient in this case.

crates/polkavm-linker/src/program_from_elf.rs
jarkkojs: By the spec what I'm doing is correct:

1. ADD32: Does +S1 relocation to the site.
2. SUB32: Does -S2 relocation to the site.

That results S1 - S2. This is exactly how those relocation accumulate to the site.

I don't know what is happening with Doom but this is what spec says.

jarkkojs: I'm still waiting for answer about doom.

jarkkojs: Screenshot From 2025-01-27 19-05-46

We can clearly see from about and also order (ADD32 first and SUB32 follows it) the signs i.e. formally `(+S1) + (-+S2) = S1 - S2`.

jarkkojs: It's possible that I have something mistake in my logic as RISC-V relocations in particular are not my area of expertise so thus I fully opened up my deduction change. So please take time review it and point out where it goes wrong,. I'm not trying to claim here that "I'm right". I'd like to know where there might be a wrong step taken.

For the same reason I asked about doom test.

jarkkojs: I can comment my logic but I cannot comment on evidence as you did not explicitly point out the evidence. That said, I assume that we are talking about the doom test.

jarkkojs: "Okay, so the offset is correct-ish, but it has the wrong sign."
A comment by @koute in:
https://github.com/paritytech/polkavm/issues/247#issuecomment-2602325358

jarkkojs: I'd also need a comment on whether there is "a test" or "tests" (plural) failing. I see in the CI run "a test" saying but your comment contradicts that.

jarkkojs: @koute thanks for the feedback :-) I'll have now new ideas how to approach this! All good.

jarkkojs:
I wrote a Python script that discovers and computes table relocations: https://gist.github.com/jarkkojs/1f64ab5b1c92deec7d75b23504f7d890

For the binary stored in test-data:

```
$ python riscv64_table_relocs.py program-for-bug_64.elf
0x000008a7       176             
0x000008e2       148             
0x00000908       148             
0x00000950       8               
0x00024a23       354             
0x00024aa3       192             
0x00024ac1       192             
0x00024b0e       134             
0x00024b35       134             
0x00024b53       110             
0x00024b7a       22              
0x00024da8       66              
0x00024e0c       28              
0x00024e35       28              
0x00024e94       40              
```

I'll check if these match the Rust code next.

PR #261, opened 2025-01-28
polkavm-linker: Fix multiple issues reported by the fuzzer

aman4150:
The fuzzer reported several valid bugs, this patch takes care of them.

Following categories of bugs were reported:

1. RegKind with a zero register.
```
[
    Reg {
        kind: ReverseByte,
        dst: A2,
        src: Zero,
    },
]
```

2. apply() does not implement Rem32AndSignExtend and RemUnsigned32AndSignExtend.
```
[
    RegReg {
        kind: RemUnsigned32AndSignExtend,
        dst: A2,
        src1: T0,
        src2: Zero,
    },
]
```

3. apply() does not implement Div32AndSignExtend and DivUnsigned32AndSignExtend/
```
[
    RegReg {
        kind: DivUnsigned32AndSignExtend,
        dst: T1,
        src1: T2,
        src2: Zero,
    },
]
```

4. apply() does not implement Rem*/Div* Zero src1 register case.
```
[
    RegReg {
        kind: Rem64,
        dst: S0,
        src1: Zero,
        src2: A3,
    },
]
```

### DIFF
--- a/crates/polkavm-linker/src/program_from_elf.rs
+++ b/crates/polkavm-linker/src/program_from_elf.rs
@@ -2082,11 +2082,21 @@ where
                 return Ok(());
             };
 
-            let Some(src) = cast_reg_non_zero(src)? else {
-                return Err(ProgramFromElfError::other("found an invalid instruction"));
-            };
-
             use crate::riscv::RegKind as K;
+
+            let Some(src) = cast_reg_non_zero(src)? else {
+                let imm = match kind {
+                    K::CountLeadingZeroBits32 | K::CountTrailingZeroBits32 => 32,
+                    K::CountLeadingZeroBits64 | K::CountTrailingZeroBits64 => 64,
+                    K::CountSetBits32 | K::CountSetBits64 => 0,
+                    K::ReverseByte => 0,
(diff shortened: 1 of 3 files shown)


jarkkojs (approved):

crates/polkavm-linker/src/program_from_elf.rs
```diff
@@ -2082,12 +2082,26 @@ where
...
+                    K::SignExtend8 | K::SignExtend16 | K::ZeroExtend16 => 0,
+                };
+
+                emit(InstExt::Basic(BasicInst::LoadImmediate { dst, imm }));
+                return Ok(());
+            }
+
             let Some(src) = cast_reg_non_zero(src)? else {
                 return Err(ProgramFromElfError::other("found an invalid instruction"));
             };
```
koute, line 2103: The error below can now never trigger, so it'd be cleaner to write this like this:

```rust
let Some(src) = cast_reg_non_zero(src)? else {
    let imm = match kind {
        ...
    };
    ....
};
```

koute (approved):
LGTM once the comment is addressed.

PR #265, opened 2025-02-05
polkavm-linker: Fix R_RISCV_ADD* && R_RISCV_SUB* relocation

aman4150:
We do not calculate the final offset correctly, because range value is reversed.

Also add a test to verify relocation is working properly.

Fixes: #247


crates/polkavm/src/tests.rs
```diff
@@ -2255,14 +2250,24 @@ impl TestInstance {
 }
 
 fn test_blob_basic_test(config: Config, optimize: bool, is_64_bit: bool) {
-    let mut i = TestInstance::new(&config, optimize, is_64_bit);
+    let elf = if is_64_bit {
+        polkavm_test_data::TEST_BLOB_64
+    } else {
+        polkavm_test_data::TEST_BLOB_32
+    };
```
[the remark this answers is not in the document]
aman4150: In my defense, I saw this ugliness after rebase. Will fix it.

crates/polkavm/src/tests.rs
```diff
@@ -2169,14 +2169,9 @@ struct TestInstance {
 }
 
 impl TestInstance {
-    fn new(config: &Config, optimize: bool, is_64_bit: bool) -> Self {
+    fn new(config: &Config, elf: &'static [u8], optimize: bool, is_64_bit: bool) -> Self {
```
koute, line 2172: Refactor this so that it doesn't take the `is_64_bit` argument but figures it out from the ELF file.

Feel free to do in this PR or another PR, up to you.

koute (approved):

crates/polkavm/src/tests.rs
aman4150: I will do it as a part of a different PR. I'll include _strip: is_64_bit_ fix in that too.

PR #267, opened 2025-02-10
polkavm-linker: Remove is_64_bit parameter from TestInstance

aman4150:
```
...because it is kinda redundant. We can query this from the parsed blob or polkavm instance.

while at it, fix strip: is_64_bit.
```

### DIFF
--- a/crates/polkavm/src/tests.rs
+++ b/crates/polkavm/src/tests.rs
@@ -2169,9 +2169,9 @@ struct TestInstance {
 }
 
 impl TestInstance {
-    fn new(config: &Config, elf: &'static [u8], optimize: bool, is_64_bit: bool) -> Self {
+    fn new(config: &Config, elf: &'static [u8], optimize: bool) -> Self {
         let _ = env_logger::try_init();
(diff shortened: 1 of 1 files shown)


crates/polkavm/src/tests.rs
```diff
@@ -2169,9 +2169,9 @@ struct TestInstance {
 }
 
 impl TestInstance {
-    fn new(config: &Config, elf: &'static [u8], optimize: bool, is_64_bit: bool) -> Self {
+    fn new(config: &Config, elf: &'static [u8], optimize: bool) -> Self {
         let _ = env_logger::try_init();
-        let blob = get_blob_impl(optimize, is_64_bit, false, elf);
+        let blob = get_blob_impl(optimize, true, false, elf);
```
koute, line 2174: I don't remember why I made it strip, but we should probably make it never strip now instead of always.

aman4150: I see, okay

koute (approved):

Issue #274, opened 2025-02-19
(linker) failed to process DWARF: unsupported value for DW_AT_high_pc


EclesioMeloJunior:
### Description

Hello everyone! I am doing some experiments with polkavm and gcc and I am trying to compile the following C program to PVM

```c
#include <stdio.h>

void sum(int a, int b) {
  int c = a + b;
}

#ifdef __riscv

#include "polkavm_guest.h"

POLKAVM_EXPORT(void, sum, int, int);

#endif
```

here is the command to get the ELF file

```shell
/usr/local/riscv-unknown-elf/bin/riscv-unknown-elf-gcc-14.2.0 -march=rv64ema -mabi=lp64e -nostdlib -nodefaultlibs -fpie \
    -Wl,--emit-relocs \
    -Wl,--unique \
    -Wl,--no-relax \
    -mrelax \
    -fpic \
    -fPIE \
    -ffast-math \
    -gdwarf-5 \
    -g3 \
    -O3 \
    -Wno-implicit-int \
    -Wno-int-conversion \
    -Wno-shift-op-parentheses \
    -Wno-visibility \
    -Wno-absolute-value \
    -Wno-pointer-sign \
    -Wno-string-plus-int \
    -fno-exceptions \
    -fno-rtti \
    -D_GNU_SOURCE \
    -o c2pvm.elf \
    ./c-example/example.c
```
but then when I call

```
polkatool link --output c2pvm.polkavm ./c2pvm.elf
```

I get the following error:

```
ERROR: failed to link "./c2pvm.elf": failed to process DWARF: unsupported value for DW_AT_high_pc: AttributeValue { value: Data8(0), offset: None }
```

I've read issue https://github.com/paritytech/polkavm/issues/208 and removed the `-Wl,--relocatable` but the problem persists.

After dumping the generated ELF file (here is the gist of the dump: [c2pvm.dump.txt](https://gist.github.com/EclesioMeloJunior/f5f723c42f49aca3050faeef8584c743#file-c2pvm-dump-txt)):

```shell 
riscv-unknown-elf-objdump -D -W ./c2pvm.elf > ./c2pvm.dump.txt
```

and looking specifically for `DW_AT_high_pc` there is 6 occurrences and in only one its value is 0:

```
<1><fd>: Abbrev Number: 10 (DW_TAG_subprogram)
    <fe>   DW_AT_name        : (indirect string, offset: 0x514f): polkavm_export_dummy0
[...]

PR #275, opened 2025-02-19
support `DW_AT_high_pc` when its data type is data8

EclesioMeloJunior:
### Description
- Add an extra arm when `gimli::DW_AT_high_pc` is of type `Data8`

### Issue
- https://github.com/paritytech/polkavm/issues/274


koute:
Thanks!

Issue #279, opened 2025-02-26
Failure to link due to unsupported instruction `Add64`


cmichi:
I've got a Rust program that builds fine, but during `polkavm_linker::program_from_elf` I get this error:

```
R_RISCV_LO12_I for an unsupported instruction: 0x2118452243 (RegImm { kind: Add64, dst: S0, src: A0, imm: 2020 }) (at <section #11+16 ('.text': '_ZN3ink7codegen8dispatch9execution12deny_payment17h59580f872247d6ebE'+16)>)
```

The respective program is built using
```bash
cargo build 
  --target=path/to/riscv64emac-unknown-none-polkavm.json 
  --release 
  -Zbuild-std=core,alloc 
  -no-default-features 
  -Zbuild-std-features=panic_immediate_abort
```

I tried different recent Rust versions, that didn't have an impact. It's the same with `polkavm` 0.21 and the latest `polkavm/master`.

I checked [the origin of the error](https://github.com/paritytech/polkavm/blob/master/crates/polkavm-linker/src/program_from_elf.rs#L8502-L8507) and it's due to an unimplemented `match` arm for `Add64`. 

I've put the binary [here](https://github.com/use-ink/polkavm/raw/df14a803751a45848e60f3b32827fdf94adafed7/examples/trigger-bug/reproducer).

You'll get a panic when attempting to link it: 

```rust
// $ INPUT=/path/to/reproduce-binary cargo run

fn main() {
    let input_path = std::env::var("INPUT").expect("no INPUT in env");
    let config = polkavm_linker::Config::default();
    let orig = std::fs::read(input_path).expect("Failed to read {input_path:?}");
    let _linked = polkavm_linker::program_from_elf(config, orig.as_ref())
        .unwrap();
}
```

The program is complex, and breaking it down to a more easily digestible reproducer will be complex. So before doing it, I wanted to create this issue to see if it already gets you somewhere. Alternatively, I can also provide you with the project to build.

aman4150:
It is a missing relocation handling of R_RISCV_LO12_I with Add64 instruction. I'll post a fix.

Thanks for reporting this @cmichi!

cmichi:
@aman4150 Thanks for the quick reaction, @aman4150! It's really helpful for us to have such a quick turnaround cycle.

cmichi:
[...]

Issue #277, opened 2025-02-24
Program traps if compiled without `--release`, but succeeds with `--release`


cmichi:
I ran into an issue whereby a program can be successfully interpreted if compiled with `cargo build --release`, but traps if compiled without `--release`.

These are the two errors that appear. The source code for each error is slightly different:
 
* `Store of 8 bytes to 0xfffdde10 failed! (pc = 336, cycle = 33)`
* `Indirect jump to dynamic address 4294835888: invalid (bad jump table index)`

I used the latest `polkavm/master`, but it's the same issue with the latest 0.21 release. I also tried different Rust versions, but it didn't have an effect.

I broke the code down from a bigger codebase (ink!) to create this reproducer [here](https://github.com/use-ink/polkavm/tree/cmichi-bug-trap-on-dev-compilation). 

You can reproduce the bugs like this:
```bash
cd examples/program-for-bug/
cargo +nightly build --no-default-features --target ../../crates/polkavm-linker/riscv64emac-unknown-none-polkavm.json -Zbuild-std="core,alloc"

cd ../trigger-bug/
RUST_LOG=trace INPUT=../program-for-bug/target/riscv64emac-unknown-none-polkavm/debug/program-for-bug cargo run 
```

Notice, that I didn't append `--release` for the build. You can execute the program successfully by executing the above build with `--release` (+ using the binary from the `release` folder as `INPUT`):

```bash
cd examples/program-for-bug/
cargo +nightly build --no-default-features --target ../../crates/polkavm-linker/riscv64emac-unknown-none-polkavm.json -Zbuild-std="core,alloc" --release

cd ../trigger-bug/
INPUT=../program-for-bug/target/riscv64emac-unknown-none-polkavm/release/program-for-bug cargo run 
```

aman4150:
I had a quick look at the problem,

In the debug binary we have this trampoline code:

```
00000000000111f0 <_ZN15program_for_bug17value_transferred17hd5da631992693564E>:
   111f0:	7159                	addi	sp,sp,-112
   111f2:	f486                	sd	ra,104(sp)
[...snipped...]
   11208:	01e080e7          	jalr	30(ra) # 11222 <_ZN110_$LT$program_for_bug..value_transferred..Sym$u20$as$u20$polkavm_derive..default_abi..private..ImportSymbol$GT$10trampoline17h5610d8e1f7bcf55eE>
[...snipped...]
   1121c:	70a6                	ld	ra,104(sp)
[...]

PR #281, opened 2025-03-03
polkavm-linker: Handle relocation being stored into the Zero register

aman4150:
Compiler can generate relocations that store the value into the Zero register. We should swap it with a "nop".

Fixes: #279

### DIFF
--- a/crates/polkavm-linker/src/program_from_elf.rs
+++ b/crates/polkavm-linker/src/program_from_elf.rs
@@ -8481,11 +8481,11 @@ where
                                 src: _,
                                 imm: _,
                             } => {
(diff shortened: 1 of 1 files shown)


koute (approved):

cmichi:
@koute @aman4150 Thanks for the fix! Any chance of getting another release?

koute:
I can cut a release, although it'd be nice to also include a fix for https://github.com/paritytech/polkavm/issues/277; @aman4150 and ETA on that?

cmichi:
@aman4150 You probably missed the above message, do you have an ETA for #277?

PR #284, opened 2025-03-17
polkavm-linker: Add a return type to the import trampoline

aman4150:
and mark the return register properly.

We are seeing that in some debug builds trampoline subroutine is not inlined, which causes epilogue and prologue to be generated. Further corrupting the stack pointer and jumping to an incorrect address when returning from the subroutine.

We fix this by adding a return type to the trampoline subroutine and marking the return register properly, and therefore removing the "noreturn" attribute from the trampoline subroutine.

Fixes: #277

### DIFF
--- a/crates/polkavm-derive-impl/src/abi_support_impl/common_private.rs
+++ b/crates/polkavm-derive-impl/src/abi_support_impl/common_private.rs
@@ -118,6 +118,14 @@ pub type ReturnTy = u64;
 #[cfg(all(target_arch = "riscv64", target_feature = "e"))]
 pub type ReturnTy = u128;
 
+#[cfg(all(target_arch = "riscv32", target_feature = "e"))]
+#[repr(C, packed(1))]
+pub struct PackedReturnTy(pub u32, pub u32);
+
+#[cfg(all(target_arch = "riscv64", target_feature = "e"))]
+#[repr(C, packed(1))]
+pub struct PackedReturnTy(pub u64, pub u64);
+
 #[cfg(all(any(target_arch = "riscv32", target_arch = "riscv64"), target_feature = "e"))]
 #[inline(always)]
 pub extern fn pack_return_ty(a0: Reg, a1: Reg) -> ReturnTy {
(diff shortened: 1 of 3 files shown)


crates/polkavm-derive-impl/src/abi_support_impl/common_private.rs
```diff
@@ -154,7 +154,7 @@ impl IntoTuple for (Reg, Reg) {
 
 #[cfg(all(any(target_arch = "riscv32", target_arch = "riscv64"), target_feature = "e"))]
 pub trait ImportSymbol {
-    extern fn trampoline(a0: Reg, a1: Reg, a2: Reg, a3: Reg, a4: Reg, a5: Reg);
+    extern fn trampoline(a0: Reg, a1: Reg, a2: Reg, a3: Reg, a4: Reg, a5: Reg) -> (Reg, Reg);
```
koute, line 157: AFAIK `(Reg, Reg)` is not FFI-safe (so the ABI is not guaranteed); this should use a struct like this as return value instead:

```rust
#[repr(C)]
struct Regs2(Reg, Reg);
```

crates/polkavm-derive-impl/src/import.rs
```diff
@@ -315,43 +315,49 @@ pub fn polkavm_import(attributes: ImportBlockAttributes, input: syn::ItemForeign
...
                                     );
                                 }
+                                ((out1 as u64) << 64) | (out0 as u64)
                             }
                         }
 
                         #[cfg(target_arch = "riscv64")]
                         impl #abi_path::private::ImportSymbol for Sym {
-                            extern fn trampoline(a0: u64, a1: u64, a2: u64, a3: u64, a4: u64, a5: u64) {
+                            extern fn trampoline(a0: u64, a1: u64, a2: u64, a3: u64, a4: u64, a5: u64) -> u128 {
```
koute, line 342: `u128` is also not FFI safe; it needs to be a struct with two members.

koute (approved):
Just double-check that this fixes https://github.com/paritytech/polkavm/issues/277 (:

aman4150:
Yes, I did that already.

PR #285, opened 2025-03-18
fuzz/fuzz_polkavm: Add correctness fuzzer for the interpreter

aman4150:
We run both recompiler and interpreter side by side in the fuzzing harness. And then compare the results of both. This is a simple fuzzer that runs the interpreter on the input and checks if the result is the same as the recompiler.

### DIFF
--- a/fuzz/fuzz_targets/fuzz_polkavm.rs
+++ b/fuzz/fuzz_targets/fuzz_polkavm.rs
@@ -4,6 +4,7 @@ use arbitrary::Arbitrary;
 use libfuzzer_sys::fuzz_target;
 
 use polkavm::Engine;
+use polkavm::InterruptKind;
 use polkavm::ModuleConfig;
 use polkavm::ProgramCounter;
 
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/fuzz_polkavm.rs
```diff
@@ -532,4 +534,92 @@ fuzz_target!(|data: Vec<OperationKind>| {
...
+        module_config.set_strict(false);
+        module_config.set_gas_metering(Some(polkavm::GasMeteringKind::Sync));
+
+        let module = polkavm::Module::from_blob(&engine, &module_config, blob.clone()).unwrap();
+        let mut instance = module.instantiate().unwrap();
+        instance.set_gas(10000);
+        instance.set_next_program_counter(ProgramCounter(0));
+        instance.set_reg(Reg::RA, RETURN_TO_HOST);
+
+        instance
```
koute, line 558: Move into a function so that it's not duplicated.

koute (approved):

fuzz/fuzz_targets/fuzz_polkavm.rs
aman4150: will do this in a follow up change.

PR #290, opened 2025-04-15
ci: Run substrate node and try to produce blocks

aman4150:
Add a regression test to compile and run node staging CLI from polkadot SDK.

### DIFF
--- a/ci/jobs/build-and-run-substrate-node.sh
+++ b/ci/jobs/build-and-run-substrate-node.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+set -euo pipefail
+cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
+
+POLKAVM_CRATES_ROOT="$(pwd)/proxy-crates"
+POLKDOT_SDK_COMMIT=c45cc4efc1861f2e5b9ec151914c51dcdfa739b4
+
+LOG_FILE="output.log"
+TARGET_DIR=target/run-substrate-node
+NODE_RUN_DURATION=100
+NODE_BLOCKS_THRESHOLD=10
+
+rustup toolchain install --component=rust-src nightly-2024-11-01-x86_64-unknown-linux-gnu
+
+cd ../..
+
+mkdir -p $TARGET_DIR
+cd $TARGET_DIR
+
+if [ ! -d "polkadot-sdk" ]; then
+    git clone --depth 1 "https://github.com/paritytech/polkadot-sdk.git"
+    git -C polkadot-sdk fetch --depth=1 origin $POLKDOT_SDK_COMMIT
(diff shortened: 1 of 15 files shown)


ci/jobs/proxy-crates/polkavm021/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl core::ops::DerefMut for RawInstance {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
```
koute, line 64: These are probably not needed. Ideally this proxy crate should only be `pub use polkavm::*`, and wrappers should be added only if there's an API difference.

crates/polkavm-derive/src/lib.rs
```diff
@@ -105,3 +105,25 @@ macro_rules! min_stack_size {
...
+#[cfg(target_pointer_width = "64")]
+#[cfg(any(all(target_arch = "riscv64", target_feature = "e"), doc))]
+#[no_mangle]
+pub unsafe extern "C" fn __atomic_fetch_add_8(address: *mut u64, value: u64) -> u64 {
+    unsafe {
+        let old_value = *address;
+        *address += value;
+        old_value
+    }
+}
```
koute, line 129: This impl is non-sense; remove it. (The `__atomic_fetch_add_8` emulates 64-bit atomics so it's needed on the 32-bit target, but on 64-bit there are native 64-bit atomics already.)

ci/jobs/build-and-run-substrate-node.sh
```diff
@@ -0,0 +1,73 @@
...
+    git -C polkadot-sdk fetch --depth=1 origin $POLKDOT_SDK_COMMIT
+    git -C polkadot-sdk checkout $POLKDOT_SDK_COMMIT
+fi
+
+cd polkadot-sdk
+
+echo '[toolchain]' > rust-toolchain.toml
+echo 'channel = "nightly-2024-11-01"' >> rust-toolchain.toml
+
+if ! grep -qi "patch.crates-io" Cargo.toml; then
```
koute, line 32: You probably can do a `git checkout Cargo.toml` to get a fresh `Cargo.toml`.

ci/jobs/build-and-run-substrate-node.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo "[patch.crates-io]" >> Cargo.toml
```
koute, line 33: Add an extra echo before this so that it appends an extra newline, in case someone will forget to add a trailing newline in polkadot-sdk's `Cargo.toml` in the future.

koute (approved):

PR #298, opened 2025-05-19
sandbox: Fix generic sandbox

aman4150:
We have been ignoring the generic sandbox for a while now. This commit fixes the generic sandbox by implementing the updated Sandbox trait and removing the old code.

Sandbox is used to provide isolation and interface to execute compiler generated native code directly on the CPU. Generic Sandbox unlike Linux Sandbox directly runs code inside Sandbox process by modifying process memory and jumping directly to the guest code.

We implemented a bunch of interface but still some things are missing, such as gas metering, aux data access and dynamic paging.

However, now we can use the generic sandbox for testing and it is able to run 90% of our tests.

**Testing**
test result: FAILED. 573 passed; 55 failed; 0 ignored; 0 measured; 2529 filtered out; finished in 12.55s

### DIFF
--- a/crates/polkavm/src/sandbox/generic.rs
+++ b/crates/polkavm/src/sandbox/generic.rs
@@ -1,26 +1,27 @@
 #![allow(clippy::manual_range_contains)]
 
 use polkavm_common::{
-    error::{ExecutionError, Trap},
+    cast::cast,
     program::Reg,
-    utils::{align_to_next_page_usize, byte_slice_init, AsUninitSliceMut},
+    utils::{align_to_next_page_usize, byte_slice_init},
     zygote::{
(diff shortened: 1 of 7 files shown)


crates/polkavm/src/sandbox/generic.rs
```diff
@@ -450,7 +449,8 @@ struct VmCtx {
     return_address: usize,
     return_stack_pointer: usize,
 
-    gas: i64,
+    arg: AtomicU32,
+    gas: AtomicI64,
```
koute, line 453: NB: the `gas` field needs to be at *exactly* offset 0x60 since this commit: https://github.com/paritytech/polkavm/commit/b4963af5745ce70a522a83e5b784b7f32ee84af8

koute (approved):

PR #303, opened 2025-06-02
interpreter: Fix out-of-range execution ProgramCounter

aman4150:
Currently, the program counter always returns 0 when the last instruction is branch instruction and it evaluates to false. The correct behavior should be returning the faulting address.

This is also weird because the program counter returns correct value if the last instruction is something else.

Also, on re-compiler we do return correct faulting address.

This patch fixes interpreter behavior and makes it consistent with re-compiler.

### DIFF
--- a/crates/polkavm/src/interpreter.rs
+++ b/crates/polkavm/src/interpreter.rs
@@ -411,6 +411,7 @@ pub(crate) struct InterpretedInstance {
     compiled_offset: u32,
     interrupt: InterruptKind,
     step_tracing: bool,
+    unresolved_program_counter: Option<ProgramCounter>,
 }
 
 impl InterpretedInstance {
@@ -433,6 +434,7 @@ impl InterpretedInstance {
             compiled_offset: 0,
             interrupt: InterruptKind::Finished,
             step_tracing,
+            unresolved_program_counter: None,
         };
 
         instance.initialize_module();
@@ -715,8 +717,14 @@ impl InterpretedInstance {
             };
 
             self.program_counter = program_counter;
-            self.compiled_offset = self.resolve_arbitrary_jump::<DEBUG>(program_counter).unwrap_or(TARGET_OUT_OF_RANGE);
             self.next_program_counter_changed = false;
+
+            if let Some(offset) = self.resolve_arbitrary_jump::<DEBUG>(program_counter) {
+                self.compiled_offset = offset;
+            } else {
+                self.compiled_offset = TARGET_OUT_OF_RANGE;
+                self.unresolved_program_counter = Some(program_counter);
+            }
 
             if DEBUG {
                 log::debug!("Starting execution at: {} [{}]", program_counter, self.compiled_offset);
@@ -1037,8 +1045,8 @@ impl<'a> Visitor<'a> {
         self.go_to_next_instruction()
     }
 
-    fn branch<const DEBUG: bool>(
-        &self,
+    fn branch<const DEBUG: bool, const IS_TARGET_TRUE_VALID: bool, const IS_TARGET_FALSE_VALID: bool>(
+        &mut self,
         s1: impl IntoRegImm,
         s2: impl IntoRegImm,
         target_true: Target,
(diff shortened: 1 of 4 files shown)


crates/polkavm/src/tests.rs
```diff
@@ -692,6 +692,23 @@ fn simple_test(engine_config: Config) {
...
+    builder.set_code(&[asm::load_imm(A0, 1), asm::load_imm(A0, 2), asm::branch_eq_imm(RA, 0, 0)], &[]);
+
+    let blob = ProgramBlob::parse(builder.into_vec().unwrap().into()).unwrap();
+    let module = Module::from_blob(&engine, &ModuleConfig::new(), blob).unwrap();
+
+    let mut instance = module.instantiate().unwrap();
+    instance.set_reg(Reg::RA, crate::RETURN_TO_HOST);
+    instance.set_next_program_counter(ProgramCounter(0));
+    match_interrupt!(instance.run().unwrap(), InterruptKind::Trap);
+    assert_eq!(instance.program_counter(), Some(ProgramCounter(9)));
```
koute, line 709: Don't hardcode this offset. Parse the instructions to get the offset (e.g. scroll up and take a look at the previous `simple_test` where it does something like that).

crates/polkavm/src/tests.rs
```diff
@@ -3797,6 +3814,8 @@ run_tests! {
     trapping_preserves_all_registers_normal_trap
     trapping_preserves_all_registers_segfault
 
+    test_out_of_range_execution
```
koute, line 3817: `test_out_of_range_execution` -> `out_of_range_execution`

crates/polkavm/src/interpreter.rs
```diff
@@ -1035,10 +1043,21 @@ impl<'a> Visitor<'a> {
...
+        if let Some(target) = self.inner.resolve_jump::<DEBUG>(pc) {
+            Some(target)
         } else {
-            Some(target_false)
+            self.inner.unresolved_program_counter = Some(pc);
+            if branch_taken {
+                Some(TARGET_INVALID_BRANCH)
+            } else {
+                Some(TARGET_OUT_OF_RANGE)
+            }
```
koute, line 1060: Hm, I don't think we want to do this here.

The whole point of the two-stage lazy "compilation" thing we have here in the interpreter is that we can do work like resolving jumps only *once* when we "compile" a given instruction, and then for any subsequent execution that work can be skipped. That's why by default the branches initially trigger the `unresolved_*` handlers, which resolve the jumps and replace the handler in-place with one that doesn't need to do it again.

crates/polkavm/src/interpreter.rs
```diff
@@ -1035,10 +1043,24 @@ impl<'a> Visitor<'a> {
...
+            self.inner.unresolved_program_counter = Some(ProgramCounter(target));
+            if branch_taken {
+                Some(TARGET_INVALID_BRANCH)
+            } else {
+                Some(TARGET_OUT_OF_RANGE)
+            }
         } else {
-            Some(target_false)
+            Some(target)
         }
```
koute, line 1064: This is still not exactly what we want. :P

In general we'd want this handler to stay exactly the same way as it was, and instead have separate handler(s) for this special case. Again, we want to minimize the execution cost - the vast majority of branches will not need this, so they shouldn't pay the cost of handling this.

.github/workflows/rust.yml
```diff
@@ -46,6 +46,8 @@ jobs:
     - uses: actions/checkout@v4
     - name: Install clippy (base toolchain)
       run: rustup component add clippy
+    - name: Install clippy (zygote toolchain)
+      run: cd crates/polkavm-zygote && rustup component add clippy
```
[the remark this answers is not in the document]
koute: This is clippy for zygote. Zygote is Linux-only, and this is action for macOS.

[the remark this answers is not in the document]
aman4150: I see, okay.

crates/polkavm/src/interpreter.rs
aman4150: The extra penalty is testing the MSB now. Is that a lot?

I don't think there is any other way to implement this (without major refactoring, I think).

koute: You shouldn't need major refactoring; just use a separate handler for cases where the jump target is not valid, no?

koute (approved):

PR #304, opened 2025-06-03
Generic sandbox gas metering and signal handler fixes

aman4150:
```
sandbox: Fix generic sandbox signal handler
Generic sandbox does correctly store register file on receiving signal.
This is fine in most cases, since there is no recovery after getting a
signal, but some of our tests assert register values after receiving a
signal.

For instance, if we run out of gas and trigger SIGILL, none of the
register values will be stored in Vmctx.
```

```
sandbox: Fix gas metering for the generic sandbox
Generic sandbox uses a different base register, instead of r15 pointing
to Vmctx, it points to start of guest memory. Vmctx base address is
calculated by [r15-4096]. This increases the number of bits required for
offset when emitting instructions for gas metering.

As a result, we need more bytes for the gas metering instructions.

Following patch reworks gas metering logic to accommodate this change.
```

After this change, only failing test cases are related to dynamic memory support.

### DIFF
--- a/crates/polkavm/src/compiler/amd64.rs
+++ b/crates/polkavm/src/compiler/amd64.rs
@@ -868,7 +868,10 @@ where
                 self.asm.push_raw(&[0x78, 0xf9]);
             } else {
                 assert_eq!(Self::vmctx_field(S::offset_table().gas), reg_indirect(RegSize::R64, r15 - 0xfa0)); // Sanity check.
-                debug_assert!(self.asm.code_mut().ends_with(&[0x49, 0x81, 0xaf, 0x60, 0xf0, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f]));
+                debug_assert!(self
+                    .asm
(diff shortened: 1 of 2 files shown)


crates/polkavm/src/compiler/amd64.rs
```diff
@@ -202,7 +202,7 @@ where
 }
 
 const GAS_METERING_TRAP_OFFSET: u64 = 3;
-const GAS_COST_OFFSET: usize = 4;
+const GAS_COST_LINUX_SANDBOX_OFFSET: usize = 4;
```
koute, line 205: 1. There should also be a constant for the generic sandbox here.
2. You should use the `on_signal_trap` in the generic sandbox instead of copy-pasting the code. It currently uses the Linux sandbox's vmctx directly, so we'd probably want to make a trait to abstract over the vmctx-kind, implement it for both vmctxes, and use that in `on_signal_trap`. Same with `on_page_fault`.

PR #306, opened 2025-06-17
Generic sandbox dynamic paging

aman4150:
Add dynamic paging support to the generic sandbox.

Since, we do not have access to a user fault handler API (like userfaultfd on Linux), we are emulating that using mprotect and SIGSEV. We mark the paged out memory as inaccessible (protection = 0) and advice the kernel to free the memory.

We keep track of paged-in memory in sandbox page_set, and rely on SIGSEV reported address and page_set to determine if this is indeed a pagefault or invalid permission.


koute:
Doesn't run all tests yet so obviously not finished, but feel free to merge it if you want to push this piecewise.

Issue #308, opened 2025-06-20
Dynamic paging off-by-one error in Linux sandbox implementation


aman4150:
zero_memory, write_memory and is_memory_accessible are affected.

Test case:
```
diff --git a/crates/polkavm/src/tests.rs b/crates/polkavm/src/tests.rs
index 71448c7..66a9278 100644
--- a/crates/polkavm/src/tests.rs
+++ b/crates/polkavm/src/tests.rs
@@ -976,6 +976,9 @@ fn dynamic_paging_basic(mut engine_config: Config) {
     // Now handle it.
     instance.zero_memory(segfault.page_address, page_size).unwrap();
 
+    assert_eq!(instance.is_memory_accessible(0x10000, 0x4, false), true);
+    assert_eq!(instance.is_memory_accessible(0x10000 + page_size, 0x4, false), false);
+
     let segfault = expect_segfault(instance.run().unwrap());
     assert_eq!(segfault.page_address, 0x10000 + page_size);
     assert_eq!(segfault.page_size, page_size);
```

PR #312, opened 2025-06-28
linux-sandbox: Fix off-by-one errors in page_set accesses

ghost:
We have off-by-one errors in the `page_set` accesses in the Linux sandbox implementation. Current design requires `page_set`` to store inclusive page ranges, but the code was incorrectly calculating the end of the range.

This commit fixes the off-by-one errors by adjusting the end of the page range calculations to be inclusive.

Additionally, it adds a test to ensure that the memory access checks work correctly after the fix.

Fixes: #308


koute (approved):
1. Double-check that in every function you added the `- 1` the length cannot be `0` (I *think* the callers enforce this currently, but I don't remember off the top of my head.)
2. Add a `debug_assert_ne!(length, 0)` everywhere to document that we require this invariant.

PR #323, opened 2025-08-14
interpreter: Do not emit two handlers for load_imm_and_jump

ghost:
Instead we emit a single handler that does both loading the immediate and jumping to the target.

Note, it does makes this expensive, because now we have to resolve the jump every time. But this is inline to what we have for `load_imm_and_jump_indirect`.

Fixes: #318

### DIFF
--- a/crates/polkavm/src/interpreter.rs
+++ b/crates/polkavm/src/interpreter.rs
@@ -1851,6 +1851,25 @@ macro_rules! define_interpreter {
         $body
     }};
 
+    (@define $handler_name:ident $body:block $self:ident, $a0:ident: Reg, $a1:ident: u32, $a2:ident: Target) => {{
+        impl Args {
+            pub fn $handler_name(a0: impl Into<Reg>, a1: u32, a2: Target) -> Args {
+                Args {
+                    a0: a0.into().to_u32(),
+                    a1,
+                    a2,
+                    ..Args::default()
+                }
+            }
+        }
+
+        let args = $self.inner.compiled_args[cast($self.inner.compiled_offset).to_usize()];
+        let $a0 = transmute_reg(args.a0);
+        let $a1 = args.a1;
+        let $a2 = args.a2;
+        $body
+    }};
+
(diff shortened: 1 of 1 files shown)


koute:
> Note, it does makes this expensive, because now we have to resolve the jump every time. But this is inline to what we have for `load_imm_and_jump_indirect`.

We don't *have* to. :P This should work the same as it did before, i.e. only resolve the jump once when it first executes, instead of doing it each time.

crates/polkavm/src/interpreter.rs
```diff
@@ -3779,6 +3807,23 @@ define_interpreter! {
         }
     }
 
+    fn unresolved_load_imm_and_jump<const DEBUG: bool>(visitor: &mut Visitor, program_counter: ProgramCounter, ra: Reg, value: u32, jump_to: u32) -> Option<Target> {
+        if DEBUG {
+            log::trace!("[{}]: unresolved load imm and jump {jump_to}", visitor.inner.compiled_offset);
```
koute, line 3812: The log here is incomplete.

crates/polkavm/src/interpreter.rs
```diff
@@ -3779,6 +3807,23 @@ define_interpreter! {
...
+        visitor.set32::<DEBUG>(ra, value);
+        if let Some(target) = visitor.inner.resolve_jump::<DEBUG>(ProgramCounter(jump_to)) {
+            let offset = visitor.inner.compiled_offset;
+            visitor.inner.compiled_handlers[cast(offset).to_usize()] = cast_handler!(raw_handlers::load_imm_and_jump::<DEBUG>);
+            visitor.inner.compiled_args[cast(offset).to_usize()] = Args::load_imm_and_jump(ra, value, target);
+
+            Some(target)
+        } else {
+            trap_impl::<DEBUG>(visitor, program_counter)
+        }
```
koute, line 3824: Missing debug logs.

koute (approved):

koute:
LGTM

PR #324, opened 2025-08-19
interpreter: get rid of InterpretedModule

ghost:
...since ro_data and rw_data are already part of the ProgramBlob.


koute (approved):
*Technically* the `ro_data` here is still unnecessary because that is part of the RO blob too, but it is an improvement nonetheless.

If we want to absolutely minimize the memory usage we could get rid of `ro_data` too, but that is a little tricky to do because of the extra padding, so any access at the boundary would have to be special-cased. Basically:

1. any access within `0..ro_data.len() - 8` would work the same as it does now, 
2. `ro_data.len()..ro_data.len()` would have to be special-cased to first initialize an 8-byte array with zeros, copy the initial bytes, and then return a reference to that array in the get slice function,
3. `ro_data.len()..ro_data_len` would be special-cased to return zeros

But this would be pretty ugly, hm.... okay, let's maybe leave it as-is.

PR #327, opened 2025-08-24
Fuzz interpreter memory usage API

aman4150:
```
fuzz: Add fuzz test for interpreter cache size limit

We update the existing polkavm fuzzing harness to includ an optional
parameter that will set the cache size limit for the interpreter.

This will help us identify any potential issues related to bounded
cache logic.

It also found a bug where u32 multiplication could overflow when
calculating the minimum number of compiled handlers.
```

```
fuzz: Add fuzz target for ProgramBlob estimate interpreter memory usage

This commit introduces a new fuzz target, `fuzz_program_blob`, designed
to test the `ProgramBlob::estimate_interpreter_memory_usage` function.
The fuzz target generates varied inputs to ensure the function can handle
a wide range of scenarios without panicking.

Fuzzing did find an issue with integer overflow in the cache size
calculation, which has been addressed in this commit.
```

Fixes: #320

### DIFF
--- a/fuzz/fuzz_targets/fuzz_polkavm.rs
+++ b/fuzz/fuzz_targets/fuzz_polkavm.rs
@@ -7,6 +7,7 @@ use polkavm::Engine;
 use polkavm::InterruptKind;
 use polkavm::ModuleConfig;
 use polkavm::ProgramCounter;
+use polkavm::SetCacheSizeLimitArgs;
 
 use polkavm_common::program::asm;
 use polkavm_common::program::Instruction;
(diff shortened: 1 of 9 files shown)


fuzz/fuzz_targets/fuzz_polkavm.rs
```diff
@@ -582,11 +597,18 @@ fn correctness_fuzzer_harness(data: Vec<OperationKind>) {
         let interrupt_recompiler = instance_recompiler.run().unwrap();
 
         if interrupt_interpreter != interrupt_recompiler {
-            panic!("interrupt code mismatch (interpreter: {:?}, recompiler: {:?})", interrupt_interpreter, interrupt_recompiler);
+            panic!(
+                "interrupt code mismatch (interpreter: {:?}, recompiler: {:?})",
+                interrupt_interpreter, interrupt_recompiler
+            );
```
koute, line 603: This reformatting in unrelated. Is this not checked on the CI? I don't mind the reformatting, but if we're going to correct it then it should be checked on the CI so that no future PR has such reformatting changes, so please make sure it's actually checked on the CI.

koute (approved):

PR #328, opened 2025-08-24
Enable all generic sandbox tests except for memset

aman4150:
```
    ci: Enable all generic sandbox tests except known failing ones
    
    While I fix the known failing tests, enable all other generic sandbox tests to
    ensure that we catch any regressions early.
```

```
   generic: Fix pinky dynamic paging test failure
    
    Pinky dynamic paging test was failing because of GPF after handling a
    page fault. On further investigation, it was found that the temporary
    register (rcx) was not being restored properly after handling a page
    fault. This causes the load address calculation to use a wrong value
    in rcx, leading to a GPF.
    
    This problem only exists on generic sandbox because we will never page
    fault in between of using TMP_REG on Linux sandbox.
    
    with this commit, we also enable the pinky dynamic paging test in CI.
```

...will post memset update as a seperate PR later.

### DIFF
--- a/ci/jobs/build-and-test.sh
+++ b/ci/jobs/build-and-test.sh
@@ -32,10 +32,9 @@ done
 
 echo ">> cargo test (generic-sandbox)"
 cargo test --features generic-sandbox -p polkavm -- \
-    tests::compiler_generic_* \
+    tests::compiler_generic_ \
     --skip tests::compiler_generic_memset_basic \
-    --skip tests::compiler_generic_memset_with_dynamic_paging \
-    --skip tests::compiler_generic_pinky_dynamic_paging_64
+    --skip tests::compiler_generic_memset_with_dynamic_paging
 
 echo ">> cargo run (examples)"
 POLKAVM_TRACE_EXECUTION=1 POLKAVM_ALLOW_INSECURE=1 cargo run -p hello-world-host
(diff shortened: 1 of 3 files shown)


ci/jobs/build-and-test.sh
```diff
@@ -32,9 +32,9 @@ done
 
 echo ">> cargo test (generic-sandbox)"
 cargo test --features generic-sandbox -p polkavm -- \
-    tests::compiler_generic_basic_test \
-    tests::compiler_generic_simple_test \
-    tests::compiler_generic_riscv_ 
+    tests::compiler_generic_* \
```
koute, line 35: This `*` here should be unnecessary?

crates/polkavm/src/sandbox/generic.rs
```diff
@@ -383,6 +383,16 @@ unsafe extern "C" fn signal_handler(signal: c_int, info: &sys::siginfo_t, contex
...
             }
 
+            //
+            // RCX is TMP_REG, which is not mapped to any guest register.
+            // therefore we store it separately here.
+            //
+            // N.B Even though it's a volatile register, we still need to
+            // restore it to handle page faults correctly.
+            //
+            debug_assert!(polkavm_common::regmap::TMP_REG == NativeReg::rcx);
```
koute, line 393: Should be a compile-time assert.

For the future: in general anything that we can assert during the compile time should be asserted on at compile time. If you're comparing one compiled constant to another then this should never be a runtime assert.

koute (approved):
This of course still needs the `memset` to be implemented, and getting it to work at least on macos (which should be tested on the CI; the recompiler should work just fine under Rosetta)

PR #341, opened 2025-10-02
Handle Rust >= 1.91 change of `target-pointer-width` JSON type

cmichi:
With the upcoming Rust 1.91, linking with `polkavm` will fail due to:

```
error: error loading target specification: target-pointer-width: invalid type: string "64", expected u16 at line 16 column 30
```

This can already be observed in Rust nightly.

The error occurs because of https://github.com/rust-lang/rust/pull/144443, which changes the `target-pointer-width` type.

This PR raises the MSRV throughout and accounts for this type change. I've raised the MSRV of `tools/{benchtool,gastool}` as well, as those use the `polkavm` linker too.

The CI will fail for now, but once Rust 1.91 is published ([October 30, 2025](https://releases.rs/docs/1.91.0/)) it will succeed. I'm already creating the PR now to surface that this is the cause of failure to link in nightly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ version = "0.29.0"
 authors = ["Jan Bujak <jan@parity.io>", "Parity Technologies <admin@parity.io>"]
 license = "MIT/Apache-2.0"
 edition = "2021"
-rust-version = "1.86.0"
+rust-version = "1.91.0"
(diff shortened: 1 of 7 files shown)


koute:
Should be fixed now with https://github.com/paritytech/polkavm/pull/347

PR #374, opened 2026-03-30
Fix memset's inconsitencies found by fuzzer

kvpanch:
- The fuzzer identified divergency in interpreter and recompiler when
        following code was executed
    
        ```
        // All registers are zeroed
    
        1 fallthrough
        2 shift_logical_left_imm_alt_64(A0, A0, 0x8F8F030F)
        3 sub_32(SP, T0, A2)
        4 branch_less_signed(A0, A0, target=0)
        5 memset()
        6 branch_less_signed(S1, A0, target=0)
        7 store_imm_indirect_u8(RA, 0, 0)
        ```
    
        when `memset` is executed, A0 is `0xFFFFFFFF8F8F030F`, which interpreter
        truncated to `0x000000008F8F030F`, while recompiler preserved A.
    
        Later that resulted that 6 branch instruction is taken by interpreter
        and it runs out of gas, while recompiler tries to executed store which
        resulted to trap.
    
        To fix it, bookkeep `dst` register before 32-bit truncation, restore the
        register to original ptr + number of executed iterations. This prevents `A0`
        clobbering that may be used later
    
    - the other problem was due inconsistent behavior on `count`:
      recompiler truncates that register to 32-bit value, while interpreter
      keeps it as 64-bit value.
    
      To fix it, keep `count` register as 64-bit integer in recompiler

### DIFF
--- a/crates/polkavm/src/tests.rs
+++ b/crates/polkavm/src/tests.rs
@@ -4571,6 +4571,44 @@ fn memset_with_dynamic_paging(mut config: Config) {
     assert_eq!(instance.gas(), 95);
 }
 
+fn memset_preserves_a0_and_a2(config: Config) {
+    let _ = env_logger::try_init();
+
+    // Memset must not truncate A0 or A2. With count=0, memset is a no-op
+    // and both registers must pass through with their upper 32 bits intact.
+    let mut builder = ProgramBlobBuilder::new(InstructionSetKind::Latest64);
+    builder.add_export_by_basic_block(0, b"main");
+    builder.set_code(
+        &[
+            // A0 = 0xffffffffffff0000 * 0xffffffffffff0000 = 0x0000000100000000
+            asm::load_imm(Reg::A0, 0xffff0000),
+            asm::mul_64(Reg::A0, Reg::A0, Reg::A0),
+            // A2 = sign_extend(0xff08bdbd) = 0xffffffffff08bdbd
+            asm::load_imm(Reg::A2, 0xff08bdbd),
+            // Swap A2 into A3 so we can set count=0 while keeping the test value.
+            asm::move_reg(Reg::A3, Reg::A2),
+            asm::load_imm(Reg::A2, 0),
+            asm::memset(),
+            asm::ret(),
+        ],
+        &[],
+    );
+
+    let blob = ProgramBlob::parse(builder.into_vec().unwrap().into()).unwrap();
+    let engine = Engine::new(&config).unwrap();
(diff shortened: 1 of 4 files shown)


koute (commented):
Few notes:

- `memset` is an experimental instruction and isn't currently enabled "in production".
- Truncating `a0` isn't necessarily a problem. All loads and stores in PVM are defined to ignore the upper 32-bits and the address space is always 32-bit (even on a 64-bit target). So `memset` should also ignore the upper bits. The remaining question is however: do we preserve the full value of the destination pointer (with the upper bits intact), or do we truncate it? Either one is fine (whichever one is faster on the recompiler backend in preferable).
- We have to be careful here so that the program can't memset any of the recompiler's internal structures (which live in the upper 32-bits of the VM sandbox's address space), which is *probably* why I was truncating the pointer. The truncation of `count` in the recompiler *might* have been a typo, but I don't remember at this point. We should probably add a test which deliberately tries to corrupt the VM's internal structures with `memset`.

kvpanch:
@koute gentle ping

kvpanch:
@copilot resolve the merge conflicts in this pull request

crates/polkavm/src/tests.rs
```diff
@@ -4571,6 +4571,44 @@ fn memset_with_dynamic_paging(mut config: Config) {
...
+    // and both registers must pass through with their upper 32 bits intact.
+    let mut builder = ProgramBlobBuilder::new(InstructionSetKind::Latest64);
+    builder.add_export_by_basic_block(0, b"main");
+    builder.set_code(
+        &[
+            // A0 = 0xffffffffffff0000 * 0xffffffffffff0000 = 0x0000000100000000
+            asm::load_imm(Reg::A0, 0xffff0000),
+            asm::mul_64(Reg::A0, Reg::A0, Reg::A0),
+            // A2 = sign_extend(0xff08bdbd) = 0xffffffffff08bdbd
+            asm::load_imm(Reg::A2, 0xff08bdbd),
```
koute: Again, this part is unnecessarily complicated because you can just use `set_reg` to set the desired values directly instead of calculating them.

The whole `set_code` ideally should just look like this:

```
builder.set_code(&[asm::memset(), asm::ret()]);
```

There's no need to involve `mul_64`, `load_imm`, etc. and overcomplicate the test because this test is supposed to test `memset` and `memset` only.

kvpanch:
@koute gentle ping

kvpanch:
@koute gentle ping

koute (approved):

koute:
Note: change the title and make sure the description makes sense before you merge.

PR #382, opened 2026-05-06
Switch interpreter dispatch to musttail (`become`) threading

kvpanch:
Replace the central loop in `InterpretedInstance::run_impl` with direct-threaded dispatch: each handler now ends in `become dispatch(self, next_off)` and `dispatch` (inlined) tail-jumps to the next handler. This collapses the per-instruction call/ret pair into a single indirect tail jump, and moves the indirect-branch site from one shared loop to the tail of every handler so the BTB learns per-edge transition patterns.

Requires nightly + `#![feature(explicit_tail_calls)]`. The recompiler is untouched.

Measured deltas on bench-pinky / bench-prime-sieve / bench-memset / bench-minimal (host: x86_64, best-of-N runs):

| Benchmark                    | Baseline   | Patched    | Delta   |
|------------------------------|------------|------------|---------|
| runtime/memset/32            |   43.66 ms |   37.99 ms | -13.0%  |
| runtime/memset/64            |   45.02 ms |   37.40 ms | -16.9%  |
| runtime/minimal/64           |   15.60 us |   12.07 us | -22.6%  |
| runtime/pinky/32             |  128.05 ms |   97.69 ms | -23.7%  |
| runtime/pinky/64             |  122.43 ms |   91.45 ms | -25.3%  |
| runtime/prime-sieve/32       |   96.20 ms |   60.76 ms | -36.8%  |
| runtime/prime-sieve/64       |   75.67 ms |   49.90 ms | -34.1%  |
| interpreter runtime mean     |            |            | -20.3%  |
| compiler runtime mean (ctrl) |            |            |  -0.6%  |


kvpanch:
@koute gentle ping

ci/jobs/build-and-test-nightly-musttail.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -euo pipefail
+cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
+cd ../..
+
+if [ "$(cat /proc/sys/vm/unprivileged_userfaultfd)" != "1" ]; then
+    echo "1" | sudo tee /proc/sys/vm/unprivileged_userfaultfd > /dev/null
+fi
```
koute: This shouldn't be here (and it's not needed for interpreter-only tests.)

ci/jobs/build-and-test-nightly-musttail.sh
```diff
@@ -0,0 +1,18 @@
...
+
+echo ">> cargo +nightly build (polkavm, experimental-musttail)"
+cargo +nightly build -p polkavm --features experimental-musttail
+
+echo ">> cargo +nightly test (polkavm, experimental-musttail)"
+cargo +nightly test -p polkavm --features experimental-musttail
+
+echo ">> cargo +nightly test --release (polkavm, experimental-musttail)"
+cargo +nightly test --release -p polkavm --features experimental-musttail
```
koute: 1. We don't need the `build` one here.
2. This should only run interpreter tests (since that's what we're interested in testing here; the recompiler tests are already being run by other jobs).
3. This should use a pinned nightly version.

crates/polkavm/Cargo.toml
```diff
@@ -50,3 +50,6 @@ generic-sandbox = []
 
 # Internal feature for testing. DO NOT USE.
 export-internals-for-testing = []
+
+# Musttail (`become`) interpreter dispatch. Requires nightly.
+experimental-musttail = []
```
koute: A more descriptive name for the feature flag would be nice (ideally you should be able to tell what it's doing without having to read a comment; if you show a random dev `experimental-musttail` they will have no idea what it does).

crates/polkavm/src/interpreter.rs
```diff
@@ -2504,6 +2515,36 @@ struct Args {
...
+    };
+}
+
+#[cfg(not(feature = "experimental-musttail"))]
+macro_rules! tail_dispatch {
+    ($self:expr, $next_off:expr) => {
+        handler_tail::<DEBUG>($self, $next_off)
+    };
+}
+
```
koute: Okay, it seems like you can put `become` in a macro and that doesn't error out?

In which case you don't need the extra file. That is, you can just do this:

```
#[cfg(feature = "experimental-musttail")]
macro_rules! tail_call {
    ($e:expr) => {
        become $e
    };
}
```

And then when you want to use it you can just do (so you don't need any extra helper functions):

```
#[cfg(feature = "experimental-musttail")]
{
    // normal version
}

#[cfg(not(feature = "experimental-musttail"))]
{
    tail_call!(...tail call version...);
}
```

ci/jobs/build-and-test-nightly-musttail.sh
kvpanch: Tests that are executed will fail otherwise https://github.com/paritytech/polkavm/actions/runs/25462114909/job/74706643514

koute: Yes, because those are recompiler tests.

kvpanch: I see your comment down below to drop some tests.  I should be able to remove this line

kvpanch:
@koute gentle ping

crates/polkavm/src/interpreter.rs
```diff
@@ -2504,6 +2515,44 @@ struct Args {
...
+// handler returns the new offset and the central loop drives the next call.
+#[cfg(feature = "interpreter-musttail-dispatch")]
+macro_rules! tail_dispatch {
+    ($self:expr, $next_off:expr) => {
+        tail_call!(dispatch::<DEBUG>($self, $next_off))
+    };
+}
+
+#[cfg(not(feature = "interpreter-musttail-dispatch"))]
+macro_rules! tail_dispatch {
```
koute: This macro is now pointless; you can just inline this.

crates/polkavm/src/interpreter.rs
```diff
@@ -2504,6 +2515,44 @@ struct Args {
...
+#[cfg(feature = "interpreter-musttail-dispatch")]
+macro_rules! tail_call {
+    ($e:expr) => {
+        become $e
+    };
+}
+
+#[cfg(feature = "interpreter-musttail-dispatch")]
+#[inline(always)]
+fn dispatch<const DEBUG: bool>(visitor: &mut InterpretedInstance, off: Target) -> HandlerResult {
```
koute: Also this function is pointless too; you can just inline it.

crates/polkavm/src/interpreter.rs
```diff
@@ -1711,16 +1723,33 @@ impl InterpretedInstance {
             log::trace!("Implicitly resuming at: [{}]", self.next_compiled_offset);
         }
 
-        let mut offset = self.next_compiled_offset;
-        loop {
+        #[cfg(feature = "interpreter-musttail-dispatch")]
+        {
+            // Normal call, not `become`: `run_impl`'s signature doesn't match a handler's.
+            // The musttail chain runs inside the handlers and unwinds back here.
```
koute: Comment seems a little useless/confusing?

crates/polkavm/src/interpreter.rs
```diff
@@ -2949,7 +2978,22 @@ macro_rules! define_interpreter {
                 #[allow(clippy::needless_lifetimes)]
                 pub fn $handler_name<'a, $(M: $M_ty,)? $(const $const: $const_ty),+>($self: &'a mut InterpretedInstance, compiled_offset: Target) -> HandlerResult {
                     let $compiled_offset = compiled_offset;
-                    define_interpreter!(@define $handler_name $body $self $compiled_offset $($arg)*)
+                    let next_off: Target = define_interpreter!(@define $handler_name $body $self $compiled_offset $($arg)*);
+                    #[cfg(feature = "interpreter-musttail-dispatch")]
+                    {
+                        if DEBUG {
+                            $self.cycle_counter += 1;
+                        }
```
koute: This is only for debugging so it's not critical, but isn't doing this here incorrect?

Let's assume we have a program composed of only a single instruction which immediately interrupts. Without tail calls we'll have:

```
cycle_counter += 1; // in run_impl
handler();
return self.interrupt;
```

And with the tail calls we have (this implementation):

```
cycle_counter += 1; // in run_impl
handler();
cycle_counter += 1; // <-- this snippet of code
return self.interrupt;
```

koute (approved):
LGTM after comments are addressed.

Also, please quickly double-check (just to be sure) the perf impact right before merging (so that we're sure the shuffling of code didn't affect the results).

PR #394, opened 2026-07-07
Fix branch gas cost mismatch between interpreter and compiler backends

davxy:
Fable assisted... please double check as I don't know polkavm internals

What I know is that it fixes https://github.com/paritytech/polkajam/actions/runs/28850402681/job/85564028388?pr=1024 
(https://github.com/paritytech/polkajam/blob/ad8187680cbd21ca0feba2fd18fc24a4dd278311/crates/node/src/chain/exec/vm/mod.rs#L839)


crates/polkavm-common/src/simulator.rs
```diff
@@ -3230,4 +3231,91 @@ mod tests {
             ",
         )
     }
+
+    #[test]
+    fn test_branch_cost_consistent_between_parsed_and_dispatch_paths() {
```
koute: For tests what should be done:

- `crates/polkavm/tests.rs` -- put a test there which reproduces the issue; the test harness there automatically runs the tests on both the interpreter and the recompiler at the same time and cross-checks the results, so if there's a divergence it'll automatically catch it for free
- this file should only have tests which verify the calculated gas costs of blocks, in exactly the form as the existing tests (you can extend `assert_timeline` so that it runs the timeline for all basic blocks; it'll also affect other tests, but that's fine)
- (optional but nice to have) a fuzz test which compares the static dispatch table vs visit_parsing -- this should be relatively easy to do: 1) have it generate random programs, 2) write a single instruction visitor which just dumps its arguments into a Vec, 3) dispatch on that instruction visitor using both ways (parsed and static table dispatch), 4) assert that both paths produced the same vec

Cargo.toml
```diff
@@ -23,7 +23,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.35.0"
+version = "0.35.1"
```
koute: Usually when patching a single crate we don't bump all of the crates, and instead we just change the `version.workspace = true` to `version = "0.35.1` for that single crate, release it, and then revert it back to `version.workspace = true`

crates/polkavm-common/src/simulator.rs
```diff
@@ -1127,9 +1127,10 @@ where
             };
         }
 
+        // `offset + args_length + 1` is the opcode of the next instruction.
```
koute: Yes, the comment is redundant :P

crates/polkavm-common/src/program.rs
```diff
@@ -4712,7 +4712,11 @@ impl ParsedInstruction {
     where
         T: ParsingVisitor,
     {
-        self.kind.visit_parsing(self.offset.0, self.next_offset.0 - self.offset.0, visitor)
+        // NOTE: `args_length` excludes the opcode byte, matching the convention used
+        // by the static dispatch tables. Saturating because synthetic instructions
+        // (e.g. in `trap_cost`) can have `next_offset == offset`.
+        self.kind
+            .visit_parsing(self.offset.0, (self.next_offset.0 - self.offset.0).saturating_sub(1), visitor)
```
koute: Logically writing it this way makes more sense I think:

```
let args_offset = self.offset.0 + 1;
self.kind.visit_parsing(self.offset.0, self.next_offset.0.saturating_sub(args_offset), visitor)
```

(i.e. `self.offset.0` -> opcode_offset, `self.offset.0 + 1` -> args_offset, `self.next_offset.0` -> next_opcode_offset, so we logically want `next_opcode_offset - args_offset`)

The comment is also redundant.

davxy: https://github.com/paritytech/polkavm/pull/394/changes/e8ae3f00798e1096258eb24ce37e6c819d8df937

crates/polkavm-common/src/simulator.rs
davxy: https://github.com/paritytech/polkavm/pull/394/changes/e8ae3f00798e1096258eb24ce37e6c819d8df937

Cargo.toml
davxy: https://github.com/paritytech/polkavm/pull/394/changes/e8ae3f00798e1096258eb24ce37e6c819d8df937

crates/polkavm-common/src/simulator.rs
```diff
@@ -1129,7 +1129,7 @@ where
 
         if self
             .code
-            .get(cast(offset).to_usize() + cast(args_length).to_usize())
+            .get(cast(offset).to_usize() + cast(args_length).to_usize() + 1)
```
koute, line 1132: If you revert this does the test in `tests.rs` fail? If not we also need a test here in `simulator.rs` (the way I described it; note that you don't have to "manually" write out the timeline -- you can just copy-paste it from what the code generates, provided your fix actually changes the output, i.e. the test should be a regression test)

davxy: https://github.com/paritytech/polkavm/pull/394/changes/89bd1b94075566d684e4a6ece6d11fad2811d0e8

koute:
Thanks!

PR #395, opened 2026-07-07
Fix the interpreter's low-memory guard threshold to match the recompiler

### LINKED ISSUE #390: Interpreter and recompiler disagree on the low-memory guard threshold
[`interpreter.rs:2266`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/interpreter.rs#L2266) should be `page_address < 0x10000` to match the [recompiler implementation](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/sandbox/generic.rs#L1131) and the [Graypaper specification](https://github.com/gavofyork/graypaper/blob/c72d861af59475a083a10da69a506c6d411ef8f9/text/pvm.tex#L168).

Under dynamic paging a guest access to the band `[2^14, 2^16)` = `[0x4000, 0x10000)` yields a recoverable `Segfault` on the interpreter backend but a terminal `Trap` on the recompiler -- same blob, same input, different `InterruptKind`. The interpreter guards `page_address < 1024 * 16` ([`interpreter.rs:2266`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/interpreter.rs#L2266)); the recompiler guards `page_address >= 0x10000` ([`sandbox/generic.rs:1131`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/sandbox/generic.rs#L1131)).

MRE (save as a test, e.g. `crates/polkavm/tests/guard_zone.rs`):
```rust
fn run_load_on(backend: BackendKind, addr: u32) -> InterruptKind {
    let src = format!("pub @main:\n    a0 = u8 [0x{addr:x}]\n    trap\n");
    let blob = ProgramBlob::parse(assemble(Some(InstructionSetKind::JamV1), &src).unwrap()[..].into()).unwrap();
    let mut cfg = Config::new();
    cfg.set_backend(Some(backend));
    cfg.set_allow_dynamic_paging(true);
    let engine = Engine::new(&cfg).unwrap();
    let mut mc = ModuleConfig::new();
    mc.set_dynamic_paging(true);
    let module = Module::from_blob(&engine, &mc, blob).unwrap();
    let entry = module.exports().next().unwrap().program_counter();
    let mut inst = module.instantiate().unwrap();
    inst.set_next_program_counter(entry);
    inst.set_reg(Reg::RA, polkavm::RETURN_TO_HOST);
    inst.run().unwrap()
}
// run_load_on(BackendKind::Interpreter, 0x4000) vs run_load_on(BackendKind::Compiler, 0x4000)
```

```
$ cargo test -p polkavm --features generic-sandbox --test guard_zone -- --nocapture
interpreter: load [0x1000] -> Trap
interpreter: load [0x4000] -> Segfault(Segfault { page_address: 16384, page_size: 4096, is_write_protected: false })
compiler:    load [0x1000] -> Trap
compiler:    load [0x4000] -> Trap
test guard_zone_backend_crosscheck ... ok
```
(`page_address: 16384` = `0x4000`. Cross-check needs a working recompiler sandbox: `--features generic-sandbox`, or `sysctl -w vm.unprivileged_userfaultfd=1` for the Linux sandbox.)

Git commit: [`443a1af`](https://github.com/paritytech/polkavm/commit/443a1af9f6f602f0c83c80f5751f12a2af60ee6e)

dimartiro:
# Description

Under dynamic paging, the interpreter guarded the inaccessible low-memory zone with `page_address < 1024 * 16` (0x4000), while the recompiler backends (`sandbox/generic.rs`, and the Linux sandbox by way of only registering userfaultfd from 0x10000) and the Graypaper use 0x10000. A guest access in `[0x4000, 0x10000)` therefore produced a recoverable `Segfault` on the interpreter but a terminal `Trap` on the recompiler.

This changes the interpreter's threshold to `0x10000` so both backends agree, and adds a regression test asserting `Trap` at 0x1000/0x4000/0xf000 and a `Segfault` with `page_address == 0x10000` at exactly the boundary (guarding against over-correction). The test is registered via `run_tests!`, so CI also exercises the compiler backends.

Verified locally that the test fails with the old threshold (`Segfault { page_address: 16384 }` instead of `Trap`) and passes with the fix.

Closes: #390

### DIFF
--- a/crates/polkavm/src/tests.rs
+++ b/crates/polkavm/src/tests.rs
@@ -1610,6 +1610,48 @@ fn dynamic_paging_read_at_bottom_of_address_space(mut engine_config: Config) {
     assert_eq!(instance.run().unwrap(), InterruptKind::Trap);
 }
 
+fn dynamic_paging_read_below_the_guard_threshold(mut engine_config: Config) {
+    engine_config.set_allow_dynamic_paging(true);
+
+    let _ = env_logger::try_init();
+
+    let engine = Engine::new(&engine_config).unwrap();
+    let page_size = get_native_page_size() as u32;
+
+    // Runs a program which does a single 4-byte load from `address` and returns how it stopped.
+    let run_load = |address: i32| {
+        let mut builder = ProgramBlobBuilder::new(InstructionSetKind::Latest64);
+        builder.add_export_by_basic_block(0, b"main");
+        builder.set_code(&[asm::load_i32(Reg::A0, address), asm::ret()], &[]);
+
+        let blob = ProgramBlob::parse(builder.into_vec().unwrap().into()).unwrap();
+        let mut module_config = ModuleConfig::new();
+        module_config.set_page_size(page_size);
+        module_config.set_dynamic_paging(true);
+        let module = Module::from_blob(&engine, &module_config, blob).unwrap();
+        let offsets: Vec<_> = module.blob().instructions().map(|inst| inst.offset).collect();
+
+        let mut instance = module.instantiate().unwrap();
+        instance.set_reg(Reg::RA, crate::RETURN_TO_HOST);
+        instance.set_next_program_counter(offsets[0]);
+        instance.run().unwrap()
+    };
+
+    // Reads fully inside the lowest 64KB of the address space must trap instead of
+    // producing a recoverable segfault, on every backend. (See #390.)
+    for address in [0x1000, 0x4000, 0xf000] {
+        assert_eq!(run_load(address), InterruptKind::Trap);
+    }
+
+    // A read straddling the boundary (partially below 0x10000 and partially at/above it) must
+    // also trap, since its lowest byte still lands in the inaccessible zone.
+    assert_eq!(run_load(0xfffe), InterruptKind::Trap);
+
+    // ...while a read fully at 0x10000 is the first one which segfaults recoverably.
+    let segfault = expect_segfault(run_load(0x10000));
+    assert_eq!(segfault.page_address, 0x10000);
+}
+
 fn dynamic_paging_read_memory_which_is_not_paged_in(mut engine_config: Config) {
     engine_config.set_allow_dynamic_paging(true);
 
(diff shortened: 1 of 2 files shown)


crates/polkavm/src/tests.rs
```diff
@@ -1610,6 +1610,53 @@ fn dynamic_paging_read_at_bottom_of_address_space(mut engine_config: Config) {
...
+
+    let _ = env_logger::try_init();
+
+    let engine = Engine::new(&engine_config).unwrap();
+    let page_size = get_native_page_size() as u32;
+
+    // Reads from the lowest 64KB of the address space must trap instead of
+    // producing a recoverable segfault, on every backend. (See #390.)
+    for address in [0x1000, 0x4000, 0xf000] {
+        let mut builder = ProgramBlobBuilder::new(InstructionSetKind::Latest32);
```
koute: There's a high chance we're going to delete the 32-bit ISA in the future, so please use the `Latest64` instead.

crates/polkavm/src/tests.rs
```diff
@@ -1610,6 +1610,53 @@ fn dynamic_paging_read_at_bottom_of_address_space(mut engine_config: Config) {
...
+    engine_config.set_allow_dynamic_paging(true);
+
+    let _ = env_logger::try_init();
+
+    let engine = Engine::new(&engine_config).unwrap();
+    let page_size = get_native_page_size() as u32;
+
+    // Reads from the lowest 64KB of the address space must trap instead of
+    // producing a recoverable segfault, on every backend. (See #390.)
+    for address in [0x1000, 0x4000, 0xf000] {
```
koute: Also would be nice to check *on* the boundary (so same read is partially under and partially over the threshold).

crates/polkavm/src/tests.rs
dimartiro: Done!

crates/polkavm/src/tests.rs
dimartiro: Done! thks

koute:
Thanks!

PR #396, opened 2026-07-07
Clamp register nibbles 13-15 to A5 in the recompiler

### LINKED ISSUE #391: Recompiler miscompiles register nibbles 13–15 to the wrong register
PVM register operands are 4-bit fields clamped to register 12 (A5): the Graypaper decodes every register operand as [`min(12, instr mod 16)`](https://github.com/gavofyork/graypaper/blob/c72d861af59475a083a10da69a506c6d411ef8f9/text/pvm.tex#L317), and the interpreter does the same in [`RawReg::get`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm-common/src/program.rs#L31) (`if value > 12 { value = 12 }`). The x86-64 recompiler does not: [`conv_reg`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/compiler/amd64.rs#L64) indexes `REG_MAP[reg.raw_unparsed() & 0b1111]` -- the raw, unclamped nibble -- and [`REG_MAP`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/compiler/amd64.rs#L53) only fills indices 0–12, leaving 13/14/15 at the default `conv_reg_const(Reg::T2)`. So a register field encoded as nibble 13/14/15 (a valid encoding the spec clamps to A5 / reg 12) is compiled as T2 / reg 4 -- the recompiler reads/writes the wrong register.

A [`debug_assert_eq!(native_reg, conv_reg_const(reg.get()))`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/compiler/amd64.rs#L66) catches this in debug builds (it panics during `from_blob`). In a release build the assertion is gone and the miscompilation is silent.

MRE -- `load_imm a5 = 0x11223344 ; trap`, with the destination register nibble flipped 12 -> 15 (both clamp to A5). Run on a release build (debug-assertions off) so the silent miscompilation shows:
```rust
// code_and_jump_table: [jt=0, jt_sz=0, code_len=7, load_imm(0x33) reg, imm=44 33 22 11, trap(0x00), bitmask=0x41]
fn run(reg_nibble: u8) -> [u64; 13] {                       // reg_nibble: 0x0c (control) or 0x0f (bug)
    let mut parts = ProgramParts::empty(InstructionSetKind::JamV1);
    parts.code_and_jump_table = vec![0,0,7, 0x33, reg_nibble, 0x44,0x33,0x22,0x11, 0x00, 0x41][..].into();
    let blob = ProgramBlob::from_parts(parts).unwrap();
    let mut cfg = Config::new();
    cfg.set_backend(Some(BackendKind::Compiler));
    cfg.set_sandbox(Some(SandboxKind::Generic));            // generic == Linux: same conv_reg/REG_MAP
    cfg.set_allow_experimental(true); cfg.set_allow_dynamic_paging(true);
    let engine = Engine::new(&cfg).unwrap();
    let mut mc = ModuleConfig::new(); mc.set_strict(true); mc.set_dynamic_paging(true);
    let module = Module::from_blob(&engine, &mc, blob).unwrap();
    let mut inst = module.instantiate().unwrap();
    inst.set_next_program_counter(ProgramCounter(0));
    inst.run().unwrap();
    let mut r = [0u64; 13]; for (i, x) in Reg::ALL.iter().enumerate() { r[i] = inst.reg(*x); } r
}
// interpreter (BackendKind::Interpreter) writes A5 (reg 12) for BOTH nibbles — correct.
```

```
                                A5 (reg 12)   T2 (reg 4)
nibble 12 (control)  interp     0x11223344    0
                     recompiler 0x11223344    0            agree
nibble 15 (bug)      interp     0x11223344    0            value -> A5 (correct, clamped)
                     recompiler 0             0x11223344   value -> T2 (wrong register)
```

Git commit: 443a1af9f6f602f0c83c80f5751f12a2af60ee6e

dimartiro:
# Description

PVM register operands are 4-bit fields the spec clamps to register 12 (A5): the Graypaper decodes them as `min(12, nibble)` and the interpreter clamps in `RawReg::get`. The x86-64 recompiler's `conv_reg` indexes `REG_MAP` with the raw unclamped nibble, and `REG_MAP` left indices 13–15 at the placeholder `T2` — so a register nibble of 13/14/15 silently compiled to T2 instead of A5 in release builds (the `debug_assert_eq!` in `conv_reg` catches it in debug builds).

This fills `REG_MAP`'s unused indices with `A5` so the recompiler matches `RawReg::get`, extends `test_conv_reg` to cover all 16 nibble values, and adds an end-to-end test (`load_imm` with the destination register nibble flipped 12 → 15, asserting the value lands in A5 and not T2) registered via `run_tests!` so CI exercises the compiler backends.

Verified locally that `test_conv_reg` fails without the fix (T2 vs A5 native registers at the `debug_assert`) and passes with it.

Closes: #391

### DIFF
--- a/crates/polkavm/src/compiler/amd64.rs
+++ b/crates/polkavm/src/compiler/amd64.rs
@@ -49,8 +49,27 @@ impl From<i32> for RegImm {
     }
 }
 
+/// PVM register operands are 4-bit fields; the spec clamps a nibble greater than the highest
+/// register index down to that register (see `RawReg::get`). This is that clamp target.
+const OUT_OF_BOUNDS_REG: Reg = {
+    let mut index = 1;
+    let mut max_reg = Reg::ALL[0];
+    while index < Reg::ALL.len() {
+        let reg = Reg::ALL[index];
+        if reg as usize > max_reg as usize {
+            max_reg = reg;
+        }
+        index += 1;
+    }
+    assert!(max_reg as usize == Reg::A5 as usize);
+    max_reg
+};
+
 static REG_MAP: [NativeReg; 16] = {
-    let mut output = [conv_reg_const(Reg::T2); 16];
+    // Register nibbles greater than the highest register index are clamped to `OUT_OF_BOUNDS_REG`
+    // (see `RawReg::get`), so map the unused indices there too.
+    // (See https://github.com/paritytech/polkavm/issues/391.)
+    let mut output = [conv_reg_const(OUT_OF_BOUNDS_REG); 16];
     let mut index = 0;
     while index < Reg::ALL.len() {
         assert!(Reg::ALL[index] as usize == index);
@@ -72,6 +91,13 @@ fn conv_reg(reg: RawReg) -> NativeReg {
 fn test_conv_reg() {
     for reg in Reg::ALL {
         assert_eq!(conv_reg(reg.into()), conv_reg_const(reg));
+    }
+
+    // Register nibbles greater than the highest register index are clamped to `OUT_OF_BOUNDS_REG`.
+    for nibble in 13..16 {
+        let reg = polkavm_common::program::read_args_regs2(u128::from(nibble)).0;
+        assert_eq!(reg.raw_unparsed(), nibble);
+        assert_eq!(conv_reg(reg), conv_reg_const(OUT_OF_BOUNDS_REG));
     }
 }
 
--- a/crates/polkavm/src/tests.rs
+++ b/crates/polkavm/src/tests.rs
@@ -5006,6 +5006,36 @@ fn jam_validate_invalid_skip(config: Config) {
     ));
 }
 
+fn jam_reg_nibble_clamped_to_a5(config: Config) {
+    let _ = env_logger::try_init();
+    let engine = Engine::new(&config).unwrap();
+
+    // Register nibbles greater than 12 are clamped to 12 (see `RawReg::get`), so a `load_imm`
+    // whose destination nibble is 13, 14 or 15 must still write to A5 and not to T2.
+    // (See https://github.com/paritytech/polkavm/issues/391.)
+    for nibble in 13..=15 {
+        let mut builder = ProgramBlobBuilder::new(InstructionSetKind::JamV1);
+        builder.add_export_by_basic_block(0, b"main");
+        builder.set_code(&[asm::load_imm(Reg::A5, 0x12345678), asm::ret()], &[]);
+        let mut blob = ProgramBlob::parse(builder.into_vec().unwrap().into()).unwrap();
+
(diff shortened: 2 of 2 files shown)


crates/polkavm/src/compiler/amd64.rs
```diff
@@ -50,7 +50,9 @@ impl From<i32> for RegImm {
 }
 
 static REG_MAP: [NativeReg; 16] = {
-    let mut output = [conv_reg_const(Reg::T2); 16];
+    // Register nibbles greater than 12 are clamped to `A5` (see `RawReg::get`),
+    // so map them to `A5` here too. (See https://github.com/paritytech/polkavm/issues/391.)
+    let mut output = [conv_reg_const(Reg::A5); 16];
```
koute: Instead of hardcoding `conv_reg_const(Reg::A5)` here please turn that into a separate constant. Maybe something along these lines:

```
const OUT_OF_BOUNDS_REG: Reg = {
    let mut index = 1;
    let mut max_reg = Reg::ALL[0];
    while index < Reg::ALL.len() {
        let reg = Reg::ALL[index];
        if reg as usize > max_reg as usize {
            max_reg = reg;
        }
        index += 1;
    }
    assert!(reg == Reg::A5);
    reg
};
```

crates/polkavm/src/tests.rs
```diff
@@ -4969,6 +4969,33 @@ fn jam_validate_invalid_skip(config: Config) {
...
+    builder.add_export_by_basic_block(0, b"main");
+    builder.set_code(&[asm::load_imm(Reg::A5, 0x12345678), asm::ret()], &[]);
+    let mut blob = ProgramBlob::parse(builder.into_vec().unwrap().into()).unwrap();
+
+    // Rewrite `load_imm`'s destination register nibble from 12 to 15; register nibbles
+    // greater than 12 are clamped to 12, so the value must still end up in A5.
+    // (See https://github.com/paritytech/polkavm/issues/391.)
+    let mut raw_code = blob.code().to_vec();
+    assert_eq!(raw_code[1], 12);
+    raw_code[1] = 15;
```
koute: Let's put this in a loop and test `13..=15` and not just 15

crates/polkavm/src/compiler/amd64.rs
dimartiro: Done! thks

crates/polkavm/src/tests.rs
dimartiro: Done

PR #397, opened 2026-07-07
Fix an off-by-one in the recompiler's branch target bounds check

### LINKED ISSUE #392: Recompiler panics compiling a branch/jump whose target == `code_length+2` (off-by-one)
[`get_or_forward_declare_label`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/compiler.rs#L573) bounds-checks the branch/jump target against the program-counter|->label map with `>` instead of `>=`:
```rust
if program_counter > self.program_counter_to_label.len() { return None; }  // off-by-one: should be >=
let label = self.asm.forward_declare_label();
self.program_counter_to_label.insert(program_counter, label);              // panics when program_counter == len()
```
The map is sized `code_length + 2` ([`compiler.rs:193/198`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/compiler.rs#L193)), so valid indices are `0 .. code_length+1`. A target equal to the map length (`code_length+2`) passes the `>` guard and indexes one past the end in [`FlatMap::insert`](https://github.com/paritytech/polkavm/blob/443a1af9f6f602f0c83c80f5751f12a2af60ee6e/crates/polkavm/src/utils.rs#L82). The recompiler crashes inside `Module::from_blob`; the interpreter runs the same blob without issue.

MRE -- `code_and_jump_table` with `code_length = 15` (map len = 17) and a `branch_less_unsigned_imm` whose target is `17`:
```rust
// [jt=0, jt_sz=0, code_len=15, <15 code bytes incl. a branch with target 17>, bitmask=ff 48]
let blob = [0u8,0,15, 0xd5,0x14,0x0a,0x3b,0x65,0x57,0x14,0xad,0xd5,0x30,0x30,0x53,0x02,0x06,0x00, 0xff,0x48];
let mut parts = ProgramParts::empty(InstructionSetKind::JamV1);
parts.code_and_jump_table = blob[..].into();
let blob = ProgramBlob::from_parts(parts).unwrap();
// interpreter: runs fine (reaches ecalli @2)
Module::from_blob(&interp_engine, &mc, blob.clone()).unwrap();
// recompiler: panics in from_blob at utils.rs:82
Module::from_blob(&compiler_engine, &mc, blob).unwrap();
```
```
interp:     Ok  (ecalli @ pc=2)
recompiler: panic  utils.rs:82  "index out of bounds: the len is 17 but the index is 17"
```
Git commit: https://github.com/paritytech/polkavm/commit/443a1af9f6f602f0c83c80f5751f12a2af60ee6e

dimartiro:
# Description

`get_or_forward_declare_label` bounds-checks a branch/jump target against the pc→label `FlatMap` with `>` instead of `>=`. The map is sized `code_length + 2`, so a target exactly equal to the map's length passed the guard and `FlatMap::insert` indexed one past the end, panicking inside `Module::from_blob` on the recompiler backend — while the interpreter accepted the same blob.

With `>=`, the function returns `None` for that target and all call sites fall back to `invalid_jump_label`, so the branch, if taken, traps — matching the interpreter's behavior.

The regression test builds a valid JamV1 blob and patches the branch-target byte so the parsed target is exactly `code_length + 2` (asserted on the re-parsed instruction), reproducing the original panic (`index out of bounds: the len is 7 but the index is 7` in `utils.rs`) without the fix. The issue's raw MRE bytes were intentionally not used verbatim: that fuzzer blob also contains a register nibble > 12, which trips the `conv_reg` debug assert in debug builds — that is issue #391, being fixed separately.

Verified locally (generic sandbox) that the test panics without the fix and passes with it, including the runtime taken-branch → `Trap` path on interpreter and compiler.

Closes: #392

### DIFF
--- a/crates/polkavm/src/compiler.rs
+++ b/crates/polkavm/src/compiler.rs
@@ -574,7 +574,8 @@ where
         match self.program_counter_to_label.get(program_counter) {
             Some(label) => Some(label),
             None => {
-                if program_counter > self.program_counter_to_label.len() {
+                // The map's length is `code_length + 2`, so a target equal to its length is also out of range. (See #392.)
+                if program_counter >= self.program_counter_to_label.len() {
                     return None;
                 }
 
--- a/crates/polkavm/src/tests.rs
+++ b/crates/polkavm/src/tests.rs
@@ -5006,6 +5006,36 @@ fn jam_validate_invalid_skip(config: Config) {
     ));
 }
 
+fn jam_branch_target_just_past_the_end_of_code(config: Config) {
+    let _ = env_logger::try_init();
+    let engine = Engine::new(&config).unwrap();
+
+    let mut builder = ProgramBlobBuilder::new(InstructionSetKind::JamV1);
+    builder.add_export_by_basic_block(0, b"main");
+    builder.set_code(&[asm::branch_eq_imm(Reg::A0, 33, 1), asm::trap()], &[]);
+
+    // Patch the branch so that its target is exactly `code_length + 2`;
+    // compiling this used to panic on the compiler backend. (See #392.)
+    let mut blob = ProgramBlob::parse(builder.into_vec().unwrap().into()).unwrap();
+    let code_length = cast(blob.code().len()).to_u32_or_panic();
+    let instructions: Vec<_> = blob.instructions().collect();
+    let mut raw_code = blob.code().to_vec();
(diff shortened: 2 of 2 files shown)


koute (approved):
Thanks!

koute:
Needs to be rebased.

dimartiro:
> Needs to be rebased.

@koute Done! thanks

PR #400, opened 2026-07-16
Allow configuring the `LineProgram` per-region instruction limit and relax sanity check

elle-j:
## Description

`program_from_elf` can emit a debug line program that the runtime parser refuses to read. A debug build panics in the linker's own sanity check:

```
thread 'main' panicked at program_from_elf.rs:
called `Result::unwrap()` on an `Err` value: ProgramParseError(Other("found a line program with too many instructions"))
```

A release build does not panic (the check is gated on `cfg!(debug_assertions) && !config.strip`) but still emits a region no consumer can parse, so backtraces for that region fail.

The line program parser (`LineProgram::raw_run`) reads at most `INSTRUCTION_LIMIT_PER_REGION` (512) opcodes per region. For one instruction, `emit_debug_info` writes a `SetStackDepth`, then up to seven opcodes per inline frame, and only then the finish op that ends the region. So a single instruction with a deep enough inline-frame stack produces a region over the limit.

This was [reported](https://github.com/paritytech/revive/issues/574) in revive when compiling using a debug build which panicked in the polkavm linker.

## Changes

The parser's instruction limit per region is made configurable instead so the sanity check can relax it:

* `struct LineProgramConfig`
  * New config with a `Default` keeping the prior hardcoded limit.
  * Raising it to `usize::MAX` essentially disables it.
* `fn LineProgram::run_with_config()`
  * `run()` delegates to it with `LineProgramConfig::default()`.
* Sanity check in `program_from_elf`
  * Runs with the limit max raised/disabled.

(This has been updated from the previous approach of capping the frame stack per instruction as per review feedback)

### DIFF
--- a/crates/polkavm-common/src/program.rs
+++ b/crates/polkavm-common/src/program.rs
@@ -6159,6 +6159,11 @@ struct LineProgramFrame {
     column: u32,
 }
 
+/// The maximum number of line program opcodes parsed for a single region.
+///
+/// Producers of the debug line program must keep every region within this limit.
+pub const INSTRUCTION_LIMIT_PER_REGION: usize = 512;
+
 /// A line program state machine.
 pub struct LineProgram<'a> {
     entry_index: usize,
@@ -6208,9 +6213,6 @@ impl<'a> LineProgram<'a> {
         if self.is_finished {
             return Ok(None);
         }
-
-        // Put an upper limit to how many instructions we'll process.
-        const INSTRUCTION_LIMIT_PER_REGION: usize = 512;
 
         let mark_as_finished_on_drop = SetTrueOnDrop(&mut self.is_finished);
         for _ in 0..INSTRUCTION_LIMIT_PER_REGION {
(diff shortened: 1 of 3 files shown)


koute:
Thanks for the PR!

Hm... this seems somewhat brittle (due to the`MAX_LINE_PROGRAM_OPCODES_PER_FRAME` and `LINE_PROGRAM_OPCODES_PER_INSTRUCTION_OVERHEAD` hardcodes), and it's probably not the way to go here (it mangles the stack trace to keep itself under the limit).

In general I didn't expect any real-world blob to hit this limit, but since you've demonstrated a counter-example we can probably relax this check. So:

- Create `LineProgram::run_with_config` to take a `LineProgramConfig` struct, where you can pass the `INSTRUCTION_LIMIT_PER_REGION` (e.g. `instruction_limit_per_region: Option<usize>` field, or maybe even just `instruction_limit_per_region: usize` since `usize::MAX` is effectively same as `None`)
- Make `LineProgram::run` call `LineProgram::run_with_config(LineProgramConfig::default())`
- The `Default` impl for that `LineProgramConfig` should set the limit as it is now hardcoded.
- In the linker run the sanity check either with a *much* higher limit (to guard against potential infinite loops due to bugs) or just with a `None` limit.

elle-j:
Thanks for the feedback @koute! I've updated it (along with the title and description).

koute:
Thanks!

PR #409, opened 2026-07-27
bench hash added

michalkucharczyk:
Adds hashing benchmarks comparing PVM vs native execution, used to decide whether the JAM Parachain Service needs crypto host calls.

Two components added:
- `bench-hash` guest crate exporting the `sp_io::hashing` functions (blake2, keccak, sha2, twox) as parameterized `benchmark_<algo>(len, times)` entry points, 
- benchtool `bench-hash` subcommand that runs them over a size grid on the PVM blob and native libraries, printing raw per-iteration times (`process-results.sh` turns the CSV into comparison tables).

Along the way this exposed that `bench-common.rs`'s byte-at-a-time `memcpy`/`memset` shadow `compiler_builtins`' weak word-wise versions that real-world guests link. This  inflated small-input results by up to 2.4× under sync gas metering (keccak_256 @ 32 B: 1.19 µs → 497 ns). They are now gated behind a per-crate `builtins-mem` feature (default-on only for `bench-hash`, so existing benchmark numbers are unaffected).


michalkucharczyk:
results are [here](https://github.com/paritytech/polkadot-sdk/blob/mku-crypto-benchmarking/designs/parachain-service-on-jam/crypto-benchamarking.md)

PR #412, opened 2026-08-04
Resolve DW_AT_ranges through its relocation

kvpanch:
A DW_FORM_sec_offset field stores zero in a relocatable object, with the real .debug_ranges offset only in its relocation — which we discarded, so every DIE resolved to the first range list. In a call-heavy function that produced inline frame stacks over a hundred deep, exceeding the per-region line program opcode limit and making the emitted debug info unreadable.

Also read range lists to their end instead of stopping after the first entry.


koute:
Do you have an example of before and after based on a real program?

kvpanch:
> Do you have an example of before and after based on a real program?

Before it was just a crash, reported in https://github.com/paritytech/revive/issues/574.

koute:
Okay, but wasn't the crash fixed (or should I say, worked around?) by https://github.com/paritytech/polkavm/pull/400?

In which case I'd like to see a before and after comparison to show that this actually fixes the debug info (which should hopefully be easy based on a real example).

kvpanch:
> Okay, but wasn't the crash fixed (or should I say, worked around?) by #400?
> 
> In which case I'd like to see a before and after comparison to show that this actually fixes the debug info (which should hopefully be easy based on a real example).

No, looks like error was there before #400

PR #417, opened 2026-09-14
Optimize the shared wide division functions

elle-j:
## Description

This focuses on optimizing the wide division and modular functions shared by both the interpreter and recompiler, to get a more realistic and improved baseline that measurements of potential further changes can be compared against (e.g. larger design changes to the x86 emitter).

**Changes summary:**

1. Using digit-serial division ([`c8bd34a`](https://github.com/paritytech/polkavm/commit/c8bd34a67efb46b56a9c2f4124c31a40d4a2e6ec)):
    * `div`, `rem`, `addmod` and `mulmod` replace their bit-serial loops with Knuth Algorithm D over 64-bit digits.
2. Specializing division per width ([`87830d8`](https://github.com/paritytech/polkavm/commit/87830d8bd223c61e3d4bf96d6f2cf87df5e2324e)):
    * The digit step specializes per divisor width, and short operands stop paying for width they do not use.
3. Fixed jump table entry size ([`59fe671`](https://github.com/paritytech/polkavm/commit/59fe67191019efdb557caf9d486ccf33780c6950)):
    * Fixes a bug where a jump table entry could get truncated (parsed fine but trapped on the first return through it).

The sections further below 👇 elaborate a bit more on each change to alleviate reviewing.

> [!NOTE]
> **Passed end-to-end differential testing via revive**
>
> Each incremental change has passed revive's e2e differential testing suite exercising both the interpreter and the JIT/recompiler.

### Execution time

Execution time results retrieved from analysis of logs from [this benchmark CI run](https://github.com/paritytech/revive-differential-tests/actions/runs/34843291296/).

> [!NOTE]
> **Baseline Used**
>
> Baseline for the deltas are the custom ops execution time measurements from the previously presented August benchmark report. Times are per-transaction wallclock in ms.

**Summary:**

The primary affected workload:
* `ed25519`
  * Characteristic: Heavy full-width modular arithmetic
  * Interpreter: Shows 52-75% reduction in execution time
  * JIT: Shows 78-84% reduction in execution time
  * All arms except "Int Yul" get below pallet-revive's EVM backend

Mean execution time in ms. "x → y" compares "x = August report's custom ops ms" to "y = this run ms" (negative = faster; bold = outside the ±5% margin). The EVM column is the August report's EVM arm (not touched by this PR):

| Workload | EVM | Int Yul | Int NY | JIT Yul | JIT NY |
|---|---|---|---|---|---|
| ed25519 | 4.35 | 18.22 → 8.79 (**−52%**) | 12.81 → 3.25 (**−75%**) | 8.78 → 1.94 (**−78%**) | 8.06 → 1.30 (**−84%**) |

<details>
<summary><strong>📊 All measurements</strong></summary>

### Summary - All workloads

| Workload | EVM | Int Yul | Int NY | JIT Yul | JIT NY |
|---|---|---|---|---|---|
| ed25519 | 4.35 | 18.22 → 8.79 (**−52%**) | 12.81 → 3.25 (**−75%**) | 8.78 → 1.94 (**−78%**) | 8.06 → 1.30 (**−84%**) |
| sha-256-hashing | 45.24 | n/a | 27.93 → 27.91 (−0.1%) | 14.58 → 13.36 (**−8%**) | 6.47 → 5.31 (**−18%**) |
| w3s | 0.51 | 1.63 → 1.58 (−3%) | 1.24 → 1.27 (+2%) | 1.55 → 1.53 (−1%) | 1.41 → 1.37 (−3%) |
| mixed-workloads | 0.34 | 1.57 → 1.59 (+1%) | 1.02 → 1.05 (+3%) | 2.06 → 2.11 (+2%) | 2.04 → 2.10 (+3%) |
| ethereum | 0.61 | 3.42 → 3.34 (−2%) | 2.46 → 2.38 (−3%) | 3.79 → 3.75 (−1%) | — → 3.50 |
| tee-attestation | 21.45 | 63.66 → 62.24 (−2%) | 33.35 → 35.81 (**+7%**) | 25.80 → 25.50 (−1%) | 22.66 → 22.55 (−0.5%) |
| create-create2 (Case 0) | 0.56 | 1.25 → 1.22 (−2%) | 1.14 → 1.14 (0%) | 1.49 → 1.48 (−1%) | 1.47 → 1.46 (−1%) |
| erc20-transfers | 0.13 | 0.58 → 0.57 (−2%) | 0.23 → 0.24 (+4%) | 0.67 → 0.66 (−1%) | 0.67 → 0.67 (0%) |

n/a: `sha-256-hashing` has no Yul interpreter arm in either run. The ethereum JIT NY cell has no baseline because the August report has no custom-ops row for it.

The measured branch state also carried a small recompiler template for in-place immediate shifts that has been superseded by changes in the base branch applied in the [`4982529`](https://github.com/paritytech/polkavm/pull/417/commits/4982529416b6b7525d64211eb5137f2cb89257af) merge commit. The `sha-256-hashing` JIT rows are mostly attributable to that superseded template.

`Mean` and `Median` cells below show August custom-ops → this run. `Min`/`Max` are this run's.

### ed25519 (1,000 transactions per interpreter arm, 25,000 per JIT arm)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int Yul | 18.22 → 8.79 | −51.8% (2.1x) | 18.23 → 8.79 | 8.74 | 8.87 |
| PolkaVM Int NY | 12.81 → 3.25 | −74.6% (3.9x) | 12.80 → 3.24 | 3.20 | 3.34 |
| PolkaVM JIT Yul | 8.78 → 1.94 | −77.9% (4.5x) | 8.84 → 1.95 | 1.76 | 2.09 |
| PolkaVM JIT NY | 8.06 → 1.30 | −83.9% (6.2x) | 8.11 → 1.32 | 1.18 | 1.48 |

### sha-256-hashing (20,000 transactions)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int NY | 27.93 → 27.91 | −0.1% | 28.29 → 28.38 | 26.00 | 31.31 |
| PolkaVM JIT Yul | 14.58 → 13.36 | −8.4% | 14.81 → 13.44 | 12.45 | 14.27 |
| PolkaVM JIT NY | 6.47 → 5.31 | −17.9% | 6.53 → 5.36 | 4.96 | 5.63 |

### w3s (~36,000 transactions)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int Yul | 1.63 → 1.58 | −3.1% | 1.10 → 1.10 | 0.65 | 4.50 |
| PolkaVM Int NY | 1.24 → 1.27 | +2.4% | 0.89 → 0.91 | 0.50 | 3.47 |
| PolkaVM JIT Yul | 1.55 → 1.53 | −1.3% | 0.95 → 0.88 | 0.57 | 5.24 |
| PolkaVM JIT NY | 1.41 → 1.37 | −2.8% | 0.88 → 0.86 | 0.56 | 4.77 |

The August variance analysis excluded w3s from its execution-time stats because the mean is carried by a varying tail of slow transactions, so w3s means compare more loosely than the other workloads.

### mixed-workloads (125,000 transactions)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int Yul | 1.57 → 1.59 | +1.3% | 1.19 → 1.18 | 0.15 | 3.34 |
| PolkaVM Int NY | 1.02 → 1.05 | +2.9% | 0.95 → 0.95 | 0.13 | 2.27 |
| PolkaVM JIT Yul | 2.06 → 2.11 | +2.4% | 1.43 → 1.42 | 0.14 | 4.76 |
| PolkaVM JIT NY | 2.04 → 2.10 | +2.9% | 1.31 → 1.39 | 0.14 | 4.73 |

### ethereum (135,000 transactions)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int Yul | 3.42 → 3.34 | −2.3% | 4.45 → 4.63 | 0.54 | 5.56 |
| PolkaVM Int NY | 2.46 → 2.38 | −3.3% | 1.58 → 1.56 | 0.49 | 4.11 |
| PolkaVM JIT Yul | 3.79 → 3.75 | −1.1% | 2.01 → 1.96 | 0.59 | 7.53 |
| PolkaVM JIT NY | — → 3.50 | n/a | — → 1.91 | 0.60 | 7.44 |

### tee-attestation (2,000 transactions)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int Yul | 63.66 → 62.24 | −2.2% | 63.33 → 62.20 | 60.00 | 65.80 |
| PolkaVM Int NY | 33.35 → 35.81 | +7.4% | 33.27 → 35.71 | 35.18 | 37.88 |
| PolkaVM JIT Yul | 25.80 → 25.50 | −1.2% | 25.67 → 25.41 | 24.95 | 26.62 |
| PolkaVM JIT NY | 22.66 → 22.55 | −0.5% | 22.63 → 22.62 | 21.82 | 24.14 |

### create-create2 Case 0 (100,000 transactions)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int Yul | 1.25 → 1.22 | −2.4% | 1.25 → 1.21 | 1.04 | 1.41 |
| PolkaVM Int NY | 1.14 → 1.14 | 0.0% | 1.14 → 1.14 | 0.94 | 1.28 |
| PolkaVM JIT Yul | 1.49 → 1.48 | −0.7% | 1.49 → 1.48 | 1.17 | 1.67 |
| PolkaVM JIT NY | 1.47 → 1.46 | −0.7% | 1.46 → 1.46 | 1.15 | 1.68 |

### erc20-transfers (160,000 transactions)

| Platform | Mean (ms) | Δ mean | Median (ms) | Min | Max |
|---|---|---|---|---|---|
| PolkaVM Int Yul | 0.58 → 0.57 | −1.7% | 0.58 → 0.57 | 0.48 | 0.65 |
| PolkaVM Int NY | 0.23 → 0.24 | +4.3% | 0.23 → 0.23 | 0.20 | 0.28 |
| PolkaVM JIT Yul | 0.67 → 0.66 | −1.5% | 0.67 → 0.67 | 0.54 | 0.80 |
| PolkaVM JIT NY | 0.67 → 0.67 | 0.0% | 0.67 → 0.67 | 0.56 | 0.80 |

</details>

### x86-64 instruction count

**Recompiler** (dynamic x86-64 instruction counts *under qemu*, per operation including the VM crossing):

"full-width" = operands whose top limb is nonzero

| operation | x86-64 instr. before | x86-64 instr. after |
|---|---|---|
| mulmod, full-width operands | 11,644 | 1,592 (7.3x fewer) |
| ed25519 verification workload, whole run | 45.6 million | 8.5 million (5.4x fewer) |

## Change 1: Using digit-serial division

**Before:**

* `div_rem` ran a fixed 256-iteration restoring loop: shift the running remainder up one bit, pull in the next numerator bit, compare against the divisor, conditionally subtract.
* `mul_mod` ran the same shape over the 512-bit product: 512 iterations, each with a data-dependent branch. On an ed25519 verification that single function was ~78% of all executed wide-arm instructions.

**After:**

* The replacement is Knuth Algorithm D over 64-bit digits:
  * Normalize the divisor so its top bit is set, then produce one quotient *digit* per limb position instead of one quotient *bit* per bit position.
  * A full-width `mulmod` becomes 5 digit steps where it used to take 512 iterations.

**Overview of one digit step** (as [`long_division_step()`](https://github.com/paritytech/polkavm/blob/05d2f0867a2f2a5980d9b990fa055b54e6443202/crates/polkavm-common/src/wide.rs#L714) performs it):

1. Estimate the digit from the top two remainder limbs divided by the divisor's top limb.
    * That 2-by-1 division is Moller and Granlund's algorithm 4: a multiply by a reciprocal computed once per division, plus two small corrections, no hardware divide in the loop.
2. Walk it back against the divisor's second limb.
    * Knuth's Theorem 4.3.1B bounds this at two decrements.
3. Multiply-subtract the divisor times the digit out of the running remainder window.
4. Add back if that borrowed out of the top, which means the estimate was still one too high: decrement the digit, add the divisor back. (This fires with probability about 2^-64 per step.)

**References used:**

* Knuth, The Art of Computer Programming Vol. 2, §4.3.1 Algorithm D
  * The long division itself, and Theorem 4.3.1B for the two-decrement bound on the estimate correction.
* Moller and Granlund, "Improved division by invariant integers"
  * Algorithm 4 for the 2-by-1 division behind each digit estimate.

## Change 2: Specializing division per width

**Before:**

* The digit step was one function taking the divisor length at runtime.
* It stayed outlined across its call sites and kept bounds checks inside the inner loops. 

**After:**

* It is now [`long_division_digits::<DIVISOR_LENGTH>`](https://github.com/paritytech/polkavm/blob/05d2f0867a2f2a5980d9b990fa055b54e6443202/crates/polkavm-common/src/wide.rs#L689) with the length a const generic (2, 3 or 4).
* The digit step receives the span of the running remainder it acts on as a slice window instead of an index into the whole numerator, which is what leaves the digit loops with no bound left to check. The steps come out as straight-line code.
```rust
// mul_mod
// full-width modulus: one digit step per limb the product actually reaches
for step in (0..=product_length.saturating_sub(4)).rev() { ... }
```
* An 8-limb product still runs 5 steps, but a product that only reaches 5 limbs now runs 2.
* The full-width path is moved below the `product < modulus` early exit, so a small product at full width returns without dividing at all.
* `div_rem` of two single-limb operands now uses the machine's own divide: 339 counted instructions down to 134.
* The divisor's reciprocal (the precomputed approximate inverse that lets each digit estimate multiply instead of divide) was itself computed with a 128-bit division, which lowers to a `__udivti3` libcall. It now comes from a 256-entry `u16` seed table.

**References used:**

* Moller and Granlund, "Improved division by invariant integers"
  * Algorithm 3 for the table-seeded reciprocal.
* [ruint v1.20.0 `src/algorithms/div/knuth.rs`](https://github.com/alloy-rs/ruint/blob/v1.20.0/src/algorithms/div/knuth.rs), was read as a comparison
  * The table-seeded reciprocal in place of a 128-bit divide.
  * The single-limb fast path.
  * The observation that the digit step has to be monomorphic/specialized to inline.

## Change 3: Fixed jump table entry size

**Before:**

* The blob writer sized each entry by counting the bytes before the *first* zero. Any target carrying a zero below its most significant non-zero byte under-reports (e.g. 65626 = `0x01005a` reports one byte instead of three).
* The entry size is the maximum over the table, so a blob only breaks when the widest byte class is small and made entirely of under-reporters, and then its entries are written truncated.

**After:**

* Entries are now sized from the most significant byte.

### DIFF
--- a/crates/polkavm-common/src/writer.rs
+++ b/crates/polkavm-common/src/writer.rs
@@ -320,9 +320,9 @@ impl ProgramBlobBuilder {
         for &target in &self.jump_table {
             let target = target + basic_block_shift;
             let target_nth_instruction = basic_block_to_instruction_index[target as usize];
-            let offset = instructions[target_nth_instruction].position.to_le_bytes();
-            jump_table_entries.push(offset);
-            jump_table_entry_size = core::cmp::max(jump_table_entry_size, offset.iter().take_while(|&&b| b != 0).count());
+            let position = instructions[target_nth_instruction].position;
+            jump_table_entries.push(position.to_le_bytes());
+            jump_table_entry_size = core::cmp::max(jump_table_entry_size, 4 - position.leading_zeros() as usize / 8);
         }
 
         let mut output = SerializedCode {
@@ -609,7 +609,9 @@ impl<'a> Writer<'a> {
 #[cfg(test)]
 mod tests {
     use super::ProgramBlobBuilder;
-    use crate::program::{Instruction, InstructionSetKind, ProgramBlob, SECTION_OPT_METADATA_HASH};
+    use crate::program::{Instruction, InstructionSetKind, ProgramBlob, Reg, SECTION_OPT_METADATA_HASH};
+    use alloc::vec;
+    use alloc::vec::Vec;
 
     fn build_minimal(metadata_hash: Option<&[u8]>) -> ProgramBlob {
         let mut builder = ProgramBlobBuilder::new(InstructionSetKind::ReviveV1);
@@ -631,5 +633,57 @@ mod tests {
     fn metadata_hash_is_empty_when_absent() {
         let blob = build_minimal(None);
         assert!(blob.metadata_hash().is_empty());
+    }
+
+    fn build_with_jump_table(code: &[Instruction], jump_table: &[u32]) -> ProgramBlob {
+        let mut builder = ProgramBlobBuilder::new(InstructionSetKind::ReviveV1);
+        builder.set_code(code, jump_table);
+        ProgramBlob::parse(builder.to_vec().unwrap().into()).unwrap()
+    }
+
+    fn jump_table_targets(blob: &ProgramBlob) -> Vec<u32> {
+        blob.jump_table().iter().map(|target| target.0).collect()
+    }
+
+    /// Every `fallthrough` is one byte and starts a new basic block, so a basic block index is also
+    /// its byte offset and the requested jump table targets are exact.
+    fn fallthroughs(count: usize) -> Vec<Instruction> {
+        vec![Instruction::fallthrough; count]
+    }
+
+    /// 65626 = 0x01005a keeps its most significant non-zero byte above a zero byte.
+    #[test]
+    fn jump_table_target_with_an_interior_zero_byte_is_not_truncated() {
+        let blob = build_with_jump_table(&fallthroughs(70000), &[65328, 65626]);
+        assert_eq!(jump_table_targets(&blob), [65328, 65626]);
+    }
+
+    /// 256 = 0x000100 has nothing but a zero byte below its most significant one.
+    #[test]
+    fn jump_table_target_that_is_a_multiple_of_256_is_not_truncated() {
+        let blob = build_with_jump_table(&fallthroughs(300), &[100, 256]);
+        assert_eq!(jump_table_targets(&blob), [100, 256]);
+    }
+
+    /// A target past 2^24 needs all four bytes, so the reader has to accept an entry size of 4. A
+    /// basic block that far in needs over 16 MiB of code, which the ten byte padding instruction
+    /// provides without starting any basic block of its own.
+    #[test]
+    fn jump_table_target_needing_four_bytes_is_not_truncated() {
+        let padding = Instruction::store_imm_indirect_u64(Reg::A0.raw(), i32::MIN, i32::MIN);
+        let mut code = vec![padding; (1 << 24) / 10 + 1];
+        code.extend(fallthroughs(2));
+        let blob = build_with_jump_table(&code, &[1]);
+        let target = blob.code().len() as u32 - 1;
+        assert!(target >= 1 << 24, "the target must need all four bytes: {target}");
+        assert_eq!(jump_table_targets(&blob), [target]);
+    }
+
+    /// 66066 = 0x010212 has no zero byte below its most significant one, the shape which was always
+    /// sized correctly.
+    #[test]
+    fn jump_table_target_without_an_interior_zero_byte_round_trips() {
+        let blob = build_with_jump_table(&fallthroughs(70000), &[65585, 66066]);
+        assert_eq!(jump_table_targets(&blob), [65585, 66066]);
     }
 }
--- a/crates/polkavm-common/src/wide.rs
+++ b/crates/polkavm-common/src/wide.rs
@@ -28,6 +28,12 @@ impl U256 {
         Self([value as u64, fill, fill, fill])
     }
 
+    /// The low four limbs of a wider value, which is as much of it as fits.
+    #[inline]
+    fn from_low_limbs(limbs: &[u64; 8]) -> Self {
+        Self([limbs[0], limbs[1], limbs[2], limbs[3]])
+    }
+
     /// The least significant limb, which is what a narrowing conversion keeps.
     #[inline]
     pub const fn low_u64(self) -> u64 {
@@ -297,11 +303,6 @@ impl U256 {
         self.0[index / 64] & (1 << (index % 64)) != 0
     }
 
-    #[inline]
-    fn set_bit(&mut self, index: usize) {
-        self.0[index / 64] |= 1 << (index % 64);
-    }
-
     /// Unsigned division and remainder. Division by zero produces zero, as `DIV` does.
     pub fn div_rem(self, divisor: Self) -> (Self, Self) {
         if divisor.is_zero() {
@@ -312,23 +313,15 @@ impl U256 {
             return (Self::ZERO, self);
         }
 
-        // Doubling the running remainder cannot carry out of the top here, unlike in
-        // `mul_mod` where the numerator is twice as wide: after s steps the remainder is
-        // below 2^s, so it only reaches 2^255 on the last one, and the loop ends there.
-        let mut quotient = Self::ZERO;
-        let mut remainder = Self::ZERO;
-        for index in (0..256).rev() {
-            remainder = remainder.shift_left(1);
-            if self.bit(index) {
-                remainder.0[0] |= 1;
-            }
-            if !remainder.less_than(divisor) {
-                remainder = remainder.wrapping_sub(divisor);
-                quotient.set_bit(index);
-            }
+        let divisor_length = significant_limbs(&divisor.0);
+        if divisor_length == 1 {
+            let (quotient, remainder) = div_rem_by_limb(&self.0, divisor.0[0]);
+            return (Self::from_low_limbs(&quotient), Self::from_u64(remainder));
         }
 
-        (quotient, remainder)
+        let numerator_length = significant_limbs(&self.0);
+        let (quotient, remainder) = long_division(&self.0[..numerator_length], &divisor.0[..divisor_length]);
+        (Self::from_low_limbs(&quotient), remainder)
     }
 
     /// Signed division, following `SDIV`: division by zero produces zero, and the one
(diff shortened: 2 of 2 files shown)


crates/polkavm-common/src/writer.rs
```diff
@@ -320,9 +320,9 @@ impl ProgramBlobBuilder {
         for &target in &self.jump_table {
             let target = target + basic_block_shift;
             let target_nth_instruction = basic_block_to_instruction_index[target as usize];
-            let offset = instructions[target_nth_instruction].position.to_le_bytes();
-            jump_table_entries.push(offset);
-            jump_table_entry_size = core::cmp::max(jump_table_entry_size, offset.iter().take_while(|&&b| b != 0).count());
+            let position = instructions[target_nth_instruction].position;
+            jump_table_entries.push(position.to_le_bytes());
+            jump_table_entry_size = core::cmp::max(jump_table_entry_size, 4 - position.leading_zeros() as usize / 8);
```
koute, line 325: Nice catch. This fix you should submit in a separate PR on `master`.

elle-j: When I noticed this I saw that it was actually already fixed as part or your recent rework of the blob format. So master already got that fix 👍

elle-j:
@koute, yeah there's still a lot more cleanup and improvements to be done which we're planning to do incrementally onto the base `cl` branch.

* Code comments
  * These are mostly Fable generated (with manual corrections and changes here and there), but I'll go over each one again and rewrite or remove.
* `push_raw`
  * Definitely a good point regarding `push_raw` which I'll update in this PR for what this PR touches (the base branch likely has more of that which we can also update in a separate PR).
* Length equation
  * I believe the plan is to make those corresponding updates from the `cl` branch once it's a bit more mature.
* Offload to `polkavm-linker`
  * Great point, yup makes a lot of sense. For shift condition specifically, I'd prefer making such dedicated instruction changes in a follow-up (still to in-development `cl` branch), to keep the current one more isolated.

xermicus (approved):
Great work thanks Linnea! Can we only take `c8bd34a` (digit-serial Knuth D) and `87830d8` (per-width specialization, table reciprocal, single-limb fast path) and drop `59fe671` as well as the shifts?

I changed the recompiler to linearly map wide registers to YMM0..15.  The in-place shift template is moot since every constant shift is now inline in YMM and the `ae310b1` emitter functions are now redundant. Would be better to keep the shld / shrd assembler additions out of ae310b1 too (we want them for the native routines). This may also speed up the interpreter.

elle-j:
@xermicus, yeah I just saw there were updates to the base branch, makes sense to omit the shift commits made here then 👍

Regarding the jump table fix ([`59fe671`](https://github.com/paritytech/polkavm/commit/59fe67191019efdb557caf9d486ccf33780c6950)), why do you prefer excluding it? That fix (which also exists on polkavm `master`) will also allow us to enable `+unaligned-scalar-mem` in revive for the Yul pipeline (it's currently only enabled for newyork), and overall blob size measurements showed that it saved roughly another 2% on the 256-bit support implementation.

(The reason it wasn't enabled for the Yul pipeline seemed to have been due to an e2e regression of the `complex/internal_function_pointers/mixed_features_2` fixture.)

xermicus:
Yes you are right, better cherry-pick 59fe671 along with the two division commits

